PR #7, opened 2020-03-18
ec2_instance: Fix spurious error message when we lose a race

flowerysong:
##### SUMMARY

It is possible for all instances to stop matching the filters between the initial check for existing instances and the first call to find_instances() in change_instance_state().

If this happened, find_instances() would previously be called a second time with an empty list of instance IDs and no filters, which should not happen and immediately ends module execution with the error "No filters provided when they were required".

A more comprehensive fix (probably including propagating the initial list of existing instances instead of wastefully calling find_instances() again) might be possible, but I don't feel like digging that deep into the logic at the moment.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ec2_instance

### DIFF
--- a/plugins/modules/ec2_instance.py
+++ b/plugins/modules/ec2_instance.py
@@ -1556,7 +1556,9 @@ def change_instance_state(filters, desired_state, ec2=None):
         await_instances(ids=list(changed) + list(unchanged), state=desired_state)
 
     change_failed = list(to_change - changed)
-    instances = find_instances(ec2, ids=list(i['InstanceId'] for i in instances))
+
+    if instances:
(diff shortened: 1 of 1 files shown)


jillr:
We had to recreate the shippable project for this repo this week due to the renaming now syncing properly to shippable, need to close/reopen PR to trigger a new job.

jillr:
Passing `existing_matches` around from ln1775 could be an option, but as you noted the logic would take some working through.  I think what you have here seems reasonable, I'm going to ask @tremble to give it a second set of eyes though as collections has taken up all of my brain power lately and I want to be sure there's not possible effects I'm not seeing here.

PR #23, opened 2020-04-06
Move module_utils out of subdir

flowerysong:
##### SUMMARY

https://github.com/ansible-collections/amazon.aws/pull/26

##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/plugins/modules/aws_acm.py
+++ b/plugins/modules/aws_acm.py
@@ -223,8 +223,8 @@ arns:
 '''
 
 
-from ansible_collections.amazon.aws.plugins.module_utils.aws.core import AnsibleAWSModule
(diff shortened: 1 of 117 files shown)


jillr:
We should be consistent between the aws collections with where we import `camel_dict_to_snake_dict` and `snake_dict_to_camel_dict` from, I've already asked s-hertel for the backstory on importing that from `ec2.py` in the other PR so let's sort that question out there first.

flowerysong:
@jillr I was less concerned about cleaning that up in this PR, since it's not (strictly speaking) part of the directory flattening and is a less intrusive change.

jillr:
@flowerysong I modified the test requirements file on you, apologies. I'd like to get this merged this week if possible, let me know if you don't have time to work on it and would like me to fix it up.

PR #38, opened 2020-04-16
Added explicit encoding for key_data string.

### LINKED ISSUE #37: ec2_win_password fails for given key_data if module is executed with python3. It works with python2.
##### Avoiding duplicates
At the time of writing there were 11 issues raised. None of them was referring to ec2_win_password

##### Affected Branches
At the time of writing there is only 1 branch at all and it is affected by the issue.

##### SUMMARY
ec2_win_password fails for given key_data if module is executed with python3. It works with python2.

##### ISSUE TYPE
- Bug Report ("Code is not compatible with python3.")
- Feature Report ("Please make modules python3 compatible.")

##### COMPONENT NAME

ec2_win_password

##### ANSIBLE VERSION

```
Ansible 2.9.5
```

##### CONFIGURATION

```paste below

```

##### STEPS TO REPRODUCE
We have a pem file which is saved in a variable with linebreaks. 
If we have a task like this

```yaml
  - name: get admin pws for systems
     ec2_win_password:
        region: "{{ aws_region }}"
        aws_access_key: "{{ access_key }}"
        aws_secret_key: "{{ secret_key }}"
        security_token: "{{ security_token }}"
        instance_id: "{{ item }}"
        key_data: "{{ sshkeyplain }}"
        key_passphrase: "{{ passphrase }}"
        wait: yes
      no_log: true
      register: passwords
      loop: "{{ system_ids }}"
```

If you execute this module with python2 backend it works.
If you execute this module with python3 backend it fails. ("unable to parse key data")

##### EXPECTED RESULTS
I expect the key_data to be parsed correctly with python3 backend.

##### ACTUAL RESULTS

Error when executing "load_pem_private_key" in python3 plain.
```python
key = load_pem_private_key(key_data, b_key_passphrase, default_backend())
  File "<stdin>", line 1, in <module>
  File "/var/lib/awx/venv/ansible/lib/python3.6/site-packages/cryptography/hazmat/primitives/serialization/base.py", line 16, in load_pem_private_key
    return backend.load_pem_private_key(data, password)
  File "/var/lib/awx/venv/ansible/lib/python3.6/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 1089, in load_pem_private_key
    password,
  File "/var/lib/awx/venv/ansible/lib/python3.6/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 1282, in _load_key
    mem_bio = self._bytes_to_bio(data)
  File "/var/lib/awx/venv/ansible/lib/python3.6/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 473, in _bytes_to_bio
    data_ptr = self._ffi.from_buffer(data)
TypeError: from_buffer() cannot return the address of a unicode object
```

##### SOLUTION PROPOSAL

Our tests showed that explicit encoding is compatible with python2 and python3. Please verify.

```python
key = load_pem_private_key(key_data.encode("ascii"), b_key_passphrase, default_backend())
```

phospi:
##### SUMMARY
Explicit encoding for pem key variable was added and tested with:
 - python 3.6.8
 - python 2.7.5

Fixes #37 
##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
ec2_win_password

##### ADDITIONAL INFORMATION
This pull request is supposed to ensure compatibility with python2 and python3 for module ec2_win_password


phospi:
Can somebody explain to me why no one is checking or merging this PR? Is there any contribution guidance? I didn't find any. 

@jillr

briantist:
@phospi this lgtm, but I'm just a user and don't have any privs on the project. I recommend trying to ping someone in IRC in the #ansible-aws channel

briantist:
@phospi I don't know why there still has been zero response to this. In the meantime you should consider [adding a changelog](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) as you'll likely be asked to once someone looks at this.

plugins/modules/ec2_win_password.py
```diff
@@ -181,7 +181,7 @@ def main():
             module.fail_json(msg="unable to parse key file")
     elif key_data is not None and key_file is None:
         try:
-            key = load_pem_private_key(key_data, b_key_passphrase, default_backend())
+            key = load_pem_private_key(key_data.encode('ascii'), b_key_passphrase, default_backend())
```
s-hertel, line 184: It looks like `load_pem_private_key` expects a bytes like object so the key_data can just use to_bytes (like b_key_passphrase).

rickmendes: This is how I did it for b_key_passphrase:

```
b_key_passphrase = to_bytes(module.params.get('key_passphrase'), err    ors='surrogate_or_strict')
```

OMG - I haven't seen this code since 2016.

phospi:
@rickmendes @s-hertel  Would you approve this PR?

rickmendes:
@phospi I would feel better about approving this if you were using a bytes object instead of an encoded string. It may work today, but you are still passing a different object than load_pem_private_key expects.

phospi:
@rickmendes Good point. I did the modification.

I did some refactoring as well. The main reason is that it is now easier to use with testing. I added a first test case to ensure basic functionality. 

If anybody could tell me which relative path is required to make tests work in shippable, that would be nice.

rickmendes:
@phospi Thanks for making those changes. I approve 👍

rickmendes:
Re-approved 👍

phospi:
@s-hertel @jillr I think this PR made enough progress to be merged. Who can do the merge?

s-hertel:
Thanks so much @phospi for the patch!

PR #52, opened 2020-04-23
Add direct_connect_gateway_id to virtual_interface

mtraynham:
##### SUMMARY
This adds `direct_connect_gateway_id` to `aws_direct_connect_virtual_interface`.
This field is only applicable in private VIF cases (`public=False`) and is
mutually exclusive to `virtual_gateway_id`.

The initial PR for the existing ansible/ansible repo was here, https://github.com/ansible/ansible/pull/48711.  It is very similar, but lacked return documentation and had a bug that allowed `direct_connect_gateway_id` for `public=True` VIFs, which is not supported by AWS.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
aws_direct_connect_virtual_interface

### DIFF
--- a/tests/unit/modules/test_aws_direct_connect_virtual_interface.py
+++ b/tests/unit/modules/test_aws_direct_connect_virtual_interface.py
@@ -1,0 +1,223 @@
+# (c) 2017 Red Hat Inc.
+#
+# This file is part of Ansible
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+# Make coding more python3-ish
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from ansible_collections.amazon.aws.tests.unit.utils.amazon_placebo_fixtures import placeboify, maybe_sleep
+from ansible_collections.community.aws.plugins.modules import aws_direct_connect_virtual_interface
+
+
+class FakeModule(object):
+    def __init__(self, **kwargs):
+        self.params = kwargs
+
+    def fail_json(self, *args, **kwargs):
+        self.exit_args = args
+        self.exit_kwargs = kwargs
+        raise Exception("FAIL")
+
+    def exit_json(self, *args, **kwargs):
+        self.exit_args = args
+        self.exit_kwargs = kwargs
+
+
+def test_find_unique_vi_by_connection_id(placeboify, maybe_sleep):
+    client = placeboify.client("directconnect")
+    vi_id = aws_direct_connect_virtual_interface.find_unique_vi(client, "dxcon-aaaaaaaa", None, None)
+    assert vi_id == "dxvif-aaaaaaaa"
+
+
+def test_find_unique_vi_by_vi_id(placeboify, maybe_sleep):
+    client = placeboify.client("directconnect")
+    vi_id = aws_direct_connect_virtual_interface.find_unique_vi(client,
+                                                                None,
+                                                                "dxvif-aaaaaaaaa",
+                                                                None)
+    assert vi_id == "dxvif-aaaaaaaa"
+
+
+def test_find_unique_vi_by_name(placeboify, maybe_sleep):
+    client = placeboify.client("directconnect")
+    vi_id = aws_direct_connect_virtual_interface.find_unique_vi(client, None, None, "aaaaaaaa")
+    assert vi_id == "dxvif-aaaaaaaa"
+
+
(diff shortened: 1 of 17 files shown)


mtraynham:
@jillr @tremble , I've added unit tests for `aws_direct_connect_virtual_interface`.  

I had to comment out two tests which actually performed the boto3 execution for creating VIFs.  For some reason placeboify is not playing the second fixture, `directconnect.DescribeVirtualInterfaces_2.json` in both cases, and I'm not really sure why.  Maybe you two might have some feedback?

tests/unit/modules/test_aws_direct_connect_virtual_interface.py
```diff
@@ -0,0 +1,223 @@
...
+                        bgp_asn=123,
+                        authentication_key="aaaa",
+                        customer_address="169.254.0.1/30",
+                        amazon_address="169.254.0.2/30",
+                        address_type="ipv4",
+                        cidr=["10.88.0.0/30"],
+                        virtual_gateway_id="xxxx",
+                        direct_connect_gateway_id="yyyy")
+    vi = aws_direct_connect_virtual_interface.assemble_params_for_creating_vi(module.params)
+    print(vi)
```
jillr, line 122: Could you please remove print statements from the tests?

mtraynham:
@jillr Thanks for taking a look at this!  

I'm looking at the LAG tests as you mentioned, but I'm a bit unsure how that fits into these tests and maybe it's just a misunderstanding on my part.  The LAG tests seem to be setting up the dependencies for a single pytest to run multiple times with different inputs.  Rather, for the tests I wrote, I was looking to have a single boto3 function return a different response on it's second invocation.

For example, I've pasted one of the tests I have commented out is below.  The `aws_direct_connect_virtual_interface.ensure_state` function should initially call `directconnect.DescribeVirtualInterfaces_1` and that returns an empty response, which intends to create a new VIF.  Then `ensure_state` would construct the VIF, post to AWS and finally calls `client.describeVirtualInterafaces` a second time to get the newly created VIF, e.g. from the placebo response `directconnect.DescribeVirtualInterfaces_2`.

If I'm way off here, feel free to let me know 😄 

```python
def test_create_public_vi(placeboify, maybe_sleep):
    client = placeboify.client("directconnect")
    module = FakeModule(state="present",
                        id_to_associate='dxcon-aaaaaaaa',
                        virtual_interface_id=None,
                        public=True,
                        name="aaaaaaaa",
                        vlan=1,
                        bgp_asn=123,
                        authentication_key="aaaa",
                        customer_address="169.254.0.1/30",
                        amazon_address="169.254.0.2/30",
                        address_type="ipv4",
                        cidr=["10.88.0.0/30"],
                        virtual_gateway_id="xxxx",
                        direct_connect_gateway_id="yyyy")
    changed, latest_state = aws_direct_connect_virtual_interface.ensure_state(client, module)
    assert changed is True
    assert latest_state is not None
```

jillr:
@mtraynham I apologize for the delay in responding, you're right the LAG tests were unrelated.  I've had a chance to work through this a bit more and made some progress but am not yet sure exactly why that test is going wrong.  It looks like `client.describe_virtual_interfaces` is not being called the expected number of times for the code path it looks like that test should be following.  It seems like `find_unique_vi()` is not being called at ln446, leading to  `directconnect.DescribeVirtualInterfaces_1.json` is being read in to `vi_state()` which fails as there's no `vi['virtualInterfaces'][0]`.  Copying `DescribeVirtualInterfaces_2.json` to `DescribeVirtualInterfaces_1.json` so that there's at least a list element there for this then results in a failure on `assert changed is True` (did not have a chance to pin down why changed status is not getting set).

Next best steps would be to track down why the code path this test is following doesn't follow our expectations.  You may be able to use mocker.patch describe_virtual_interfaces to get a call count or set some side effects or print the stack to see where things are going.  I will try to dig into this further when I can.

jillr:
Hi @mtraynham, wanted to share the 2.10 [roadmap](https://docs.ansible.com/ansible/devel/roadmap/COLLECTIONS_2_10.html) dates with you.  Collections will have a freeze on new modules on 2020-08-18.  It would be great to get these tests in order so we could include the direct connect modules in that build possible.

mtraynham:
Thanks for the update @jillr.  I added some print statements to help me figure out the issue.  Seems retrieving the direct connect occurs 3 times in those test suites and not 2.  Therefore placeboify was I guess replaying from the start, expecting 0 VIFs to exist.

Issue #59, opened 2020-04-29
Add tags to the galaxy.yml


JPrause:
Needed for Content Collection v1.0

PR #60, opened 2020-04-29
Adding labels to the tags section

JPrause:
Fixes https://github.com/ansible-collections/community.aws/issues/59

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -6,7 +6,7 @@ authors: null
 description: null
 license: GPL-3.0-or-later
 license_file: COPYING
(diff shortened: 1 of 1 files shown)


galaxy.yml
```diff
@@ -6,7 +6,7 @@ authors: null
 description: null
 license: GPL-3.0-or-later
 license_file: COPYING
-tags: null
+tags: [community, aws, networking, cloud, amazon]
```
jillr, line 9: Need to remove networking in this collection as well, otherwise looks good.

PR #76, opened 2020-05-18
Disable smoketests 

jillr:
##### SUMMARY
Smoketests are not needed in collections, and none exist in this
collection anyway.
https://github.com/ansible-collections/overview/issues/45#issuecomment-630466482

~~`amazon.aws` will be using the pip package, so keep this one consistent.~~
We won't be using codecov-python, does not seem to be a drop-in replacement after all.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests/utils/shippable/aws.sh
tests/utils/shippable/shippable.sh

### DIFF
--- a/tests/utils/shippable/aws.sh
+++ b/tests/utils/shippable/aws.sh
@@ -13,22 +13,7 @@ target="shippable/${cloud}/group${group}/"
 
 stage="${S:-prod}"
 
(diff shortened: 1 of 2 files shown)


jillr:
@felixfontein would you be able to give this shippable.sh change a look over?

PR #91, opened 2020-06-07
Add validation to ensure ecs service is deployed successfully

raags:
##### SUMMARY

This expands the existing `delay` and `repeat` parameters to also check if a service is deployed successfully. This is done by checking the running count is equal to the desired count for the active deployment.

When the `delay` is set to 0, it'll disable this check.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ecs_service

##### ADDITIONAL INFORMATION

Currently, the ecs_service module starts a ECS service deployment but doesn't validate the deployment success. This change will optionally validate if the deployment succeeded, giving early feedback on the service status.

### DIFF
--- a/plugins/modules/ecs_service.py
+++ b/plugins/modules/ecs_service.py
@@ -72,13 +72,14 @@ options:
         type: str
     delay:
         description:
-          - The time to wait before checking that the service is available.
+          - The time to wait before checking that the service is available or
+            removed. Set to 0 to disable check.
         required: false
         default: 10
         type: int
     repeat:
         description:
-          - The number of times to check that the service is available.
+          - The number of times to check that the service is available or removed
         required: false
         default: 10
         type: int
@@ -730,8 +731,10 @@ def main():
                 update = True
 
         if not matching:
-            if not module.check_mode:
+            if module.check_mode:
+                results['changed'] = True
 
+            else:
                 role = module.params['role']
                 clientToken = module.params['client_token']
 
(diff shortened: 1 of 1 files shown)


markuman:
@raags thanks for your pr.

basically waiting for `desired_count` becomes equal `running_count` might not be sufficient in case your ecs service has got a load balancer configuration.  
in this case you must wait until `desired_count` becomes equal to `healthy targets` in the belonging target group.  
In case you update just the taskdefinition and the `desired_count` has not changed, it becomes more complicated. Because you must only count the healthy targets with the new taskdefinition.

In case you doesn't have a load balancer configuration (_you do some background processing jobs with ecs_), it might be sufficient.

You can achive something similar by using the `elb_target_group_info` modules.

```yml
- name: wait until at least one target is registered and healthy
  elb_target_group_facts:
    names:
      - "{{ SERVICE }}"
    collect_targets_health: yes
  register: TG
  retries: 10
  delay: 15
  until: TG.target_groups[0].targets_health_description | count == DESIRED_COUNT and TG.target_groups[0].targets_health_description[0].target_health.state == "healthy"
```

But as you see, it gets complicated. But it is definitely a "very nice to have" feature. 

Furthermore, consider to add a waiter to the `waiters.py` file in the `amazon.aws` collection rather than writing a loop inside your module.  
You'll find an example in this blog post by @jillr https://www.ansible.com/blog/getting-started-with-aws-ansible-module-development

PR #96, opened 2020-06-11
Collections related fixes for CI

jillr:
##### SUMMARY
We've been using galaxy to install amazon.aws in shippable, but that
doesn't really work if we aren't publising faster. Get that collection
from git so it is most up to date.

Also specify python deps needed for shippable.

And update modules deprecations; collections won't track ansible-base versions for deprecations.

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -219,161 +219,161 @@ plugin_routing:
   modules:
     aws_acm_facts:
       deprecation:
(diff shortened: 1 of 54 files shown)


jillr:
Life moves fast; multiple things are now broken in CI related to collections and related changes.  This is now by necessity the "fix everything" PR.

PR #97, opened 2020-06-12
Tylerl/global accelerator update

TylerLubeck:
##### SUMMARY
Implement the Global Accelerator create/update module for #94 

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
global_accelerator

### DIFF
--- a/plugins/modules/global_accelerator.py
+++ b/plugins/modules/global_accelerator.py
@@ -1,0 +1,309 @@
+from ansible_collections.amazon.aws.plugins.module_utils.aws.core import AnsibleAWSModule
+from ansible_collections.amazon.aws.plugins.module_utils.ec2 import (
+    AWSRetry,
+    camel_dict_to_snake_dict,
+    boto3_tag_list_to_ansible_dict,
+    compare_aws_tags,
+    ansible_dict_to_boto3_tag_list,
+)
+
+# Non-ansible imports
+try:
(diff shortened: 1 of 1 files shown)


plugins/modules/global_accelerator.py
```diff
@@ -0,0 +1,411 @@
...
+            connection.describe_accelerator
+        )(AcceleratorArn=arn)['Accelerator']
+
+    if accelerator:
+        accelerator = camel_dict_to_snake_dict(accelerator)
+
+    return accelerator
+
+
+class GlobalAccelerator(object):
```
TylerLubeck, line 157: TBH I'm mad confused about where the various files for this repo live. Based on other modules here, it seems like this class should live somewhere else. I'm happy to move it if someone can point me in a direction

jillr: Is this class needed for this module?  These methods should be able to be moved into the module and called from main() based on the action the user is requesting.

jillr (commented):
Thanks for this new module PR @TylerLubeck.  This will need an [integration test](https://docs.ansible.com/ansible/devel/dev_guide/platforms/aws_guidelines.html#integration-tests-for-aws-modules) suite to be added at test/integration/global_accelerator, which will likely also need [IAM policies](https://docs.ansible.com/ansible/devel/dev_guide/platforms/aws_guidelines.html#aws-permissions-for-integration-tests) added for our CI.

PR #118, opened 2020-06-24
sanity: Add future-import boilerplate

Akasurde:
##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/sanity_fix_future_boilerplate.yml
scripts/inventory/ec2.py
tests/integration/targets/aws_lambda/files/mini_lambda.py
tests/integration/targets/lambda_policy/files/mini_http_lambda.py
tests/integration/targets/s3_bucket_notification/files/mini_lambda.py
tests/integration/targets/sns_topic/files/sns_topic_lambda/sns_topic_lambda.py
tests/sanity/ignore-2.10.txt
tests/unit/mock/path.py
tests/unit/mock/vault_helper.py
tests/unit/mock/yaml_helper.py
tests/unit/modules/conftest.py
tests/unit/modules/utils.py


changelogs/sanity_fix_future_boilerplate.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
```
jillr, line 1: My understanding of antsibull-changelog is that we want to keep fragments under changelogs/fragments, could you move this into that directory please?  Otherwise this lgtm, thanks!

jillr:
Ah, we also need to remove the relevant lines from tests/sanity/ignore-2.9.txt and tests/sanity/ignore-2.11.txt. Thanks for this patch!

PR #217, opened 2020-09-01
Adding new module s3_metrics_configuration

vorotech:
##### SUMMARY
Manage s3 bucket metrics configuration in AWS which allows to get the CloudWatch request metrics for the objects in a bucket

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
s3_metrics_configuration

##### ADDITIONAL INFORMATION

* [Amazon S3 CloudWatch Request Metrics](https://docs.aws.amazon.com/AmazonS3/latest/dev/cloudwatch-monitoring.html#s3-request-cloudwatch-metrics)
* [Metrics configurations for buckets](https://docs.aws.amazon.com/AmazonS3/latest/dev/metrics-configurations.html)

The S3 Metrics Configuration object has a very non-trivial scheme. 
The **Filter** is optional, but when specified it accepts either single **Prefix**, single **Tag** or conjunction which is packed to **And** node with optional **Prefix** and collection of **Tags**.
In other words, I tried to reduce the complexity of the validation scheme by providing consumer a simple to use interface. That's why the code might seem clunky in some places.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<MetricsConfiguration>
   <Id>string</Id>
   <Filter>
      <And>
         <Prefix>string</Prefix>
         <Tags>
           <Tag>
              <Key>string</Key>
              <Value>string</Value>
           </Tag>
           ...
        </Tags>
      </And>
      <Prefix>string</Prefix>
      <Tag>
         <Key>string</Key>
         <Value>string</Value>
      </Tag>
   </Filter>
</MetricsConfiguration>
```


vorotech:
Trying to understand how to extend IAM permissions available for STS credentials passed to integration tests.

The current error is the following. 
```
Failed to get bucket metrics configuration: An error occurred (AccessDenied) when calling the GetBucketMetricsConfiguration operation: Access Denied
```

The required permissions are:

* s3:GetMetricsConfiguration 
* s3:PutMetricsConfiguration

vorotech:
Requires this PR to pass tests https://github.com/mattclay/aws-terminator/pull/109

vorotech:
ready_for_review

tests/integration/targets/s3_metrics_configuration/tasks/main.yml
```diff
@@ -0,0 +1,242 @@
...
+    register: result
+
+  - assert:
+      that:
+      - result is not changed
+
+  # ============================================================
+  - name: Put a metrics configuration that enables metrics for objects starting with a prefix
+    s3_metrics_configuration:
+      bucket_name: "{{ bucket_name }}"
```
jillr, line 106:
```suggestion
      bucket_name: "{{ test_bucket }}"
```
And several additional usages below

plugins/modules/s3_metrics_configuration.py
```diff
@@ -0,0 +1,246 @@
...
+            'Prefix': filter_prefix
+        }
+    elif len(filter_tags) > 0:
+        payload['Filter'] = {
+            'And': {
+                'Tags': ansible_dict_to_boto3_tag_list(filter_tags)
+            }
+        }
+    if filter_prefix is not None:
+        payload['Filter']['And']['Prefix'] = filter_prefix
```
jillr, line 130: This results in a KeyError when the previous `if` blocks falls though the `len(filter_tags) == 0 and filter_prefix is not None` condition because `payload['Filter']['And']` does not exist when making this assignment.

The test at: https://github.com/ansible-collections/community.aws/blob/e504e52dffac34d5262d621f174b2c5b95d7903d/tests/integration/targets/s3_metrics_configuration/tasks/main.yml#L104 triggers the failure.

jillr (changes requested):
Couple of comments.  Thanks very much for this new module!

plugins/modules/s3_metrics_configuration.py
vorotech: 🤦 the indentation issue

tests/integration/targets/s3_metrics_configuration/tasks/main.yml
vorotech: good spot

jillr (commented):
Thanks @vorotech. This looks pretty good to me, I've merged the policy changes and relaunched the failing tests.
@tremble would you have time to give this new module a look?  Tests are passing for me locally.

plugins/modules/s3_metrics_configuration.py
```diff
@@ -0,0 +1,246 @@
...
+    filter_tags = module.params.get('filter_tags')
+
+    try:
+        metrics_configuration = client.get_bucket_metrics_configuration(Bucket=bucket_name, Id=mc_id)
+    except is_boto3_error_code('NoSuchConfiguration'):
+        metrics_configuration = None
+    except ClientError as e:  # pylint: disable=duplicate-except
+        module.fail_json_aws(e, msg="Failed to get bucket metrics configuration")
+
+    if metrics_configuration:
```
tremble, line 170: There is already `_create_metrics_configuration()` which converts mc_id/filter_prefix/filter_tags into a dict in the same format as metrics_configuration is there a reason to compare them in a different format?

I think it should also be possible to simplify this logic a little:

```
new_configuration = _create_metrics_configuration(mc_id, filter_prefix, filter_tags)

if metrics_configuration:
    if metrics_configuration == new_configuration
        module.exit_json(changed=False)

if module.check_mode:
    module.exit_json(changed=True)

# Create and update both use the same call
<set_configuration>

module.exit_json(changed=True)
```

plugins/modules/s3_metrics_configuration.py
```diff
@@ -0,0 +1,246 @@
...
+    )
+    module = AnsibleAWSModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True
+    )
+
+    state = module.params.get('state')
+
+    try:
+        client = module.client('s3', retry_decorator=AWSRetry.exponential_backoff(retries=10, delay=3))
```
tremble, line 235: Using the retry_decorator is a really good thing to do, it's especially helpful with a busy account that might start hitting the rate limits (our CI does during 'full' runs for example).  As a quirk of how AWSRetry was originally written, even with the decorator configured you still need to explicitly request retries (It used to retry on certain 'NotFound' type errors, which wasn't the right thing to do for many calls)

Enabling the retries for any specific call is however *really* easy:
```
client.delete_bucket_metrics_configuration(aws_retry=True, ...)
```

plugins/modules/s3_metrics_configuration.py
```diff
@@ -0,0 +1,246 @@
...
+def delete_metrics_configuration(client, module):
+    changed = False
+    bucket_name = module.params.get('bucket_name')
+    mc_id = module.params.get('id')
+
+    try:
+        client.get_bucket_metrics_configuration(Bucket=bucket_name, Id=mc_id)
+    except is_boto3_error_code('NoSuchConfiguration'):
+        module.exit_json(changed=changed)
+    except ClientError as e:  # pylint: disable=duplicate-except
```
tremble, line 206:
```suggestion
    except (BotoCoreError, ClientError) as e:  # pylint: disable=duplicate-except
```

It's an edge case, but it's possible to throw BotoCoreErrror here (for example if S3's API was down in the region for some reason)

plugins/modules/s3_metrics_configuration.py
```diff
@@ -0,0 +1,246 @@
...
+        module.fail_json_aws(e, msg="Failed to get bucket metrics configuration")
+
+    try:
+        if not module.check_mode:
+            client.delete_bucket_metrics_configuration(Bucket=bucket_name, Id=mc_id)
+        changed = True
+    except (BotoCoreError, ClientError) as e:
```
tremble, line 213:
```suggestion
    except is_boto3_error_code('NoSuchConfiguration'):
        module.exit_json(changed=changed)
    except (BotoCoreError, ClientError) as e:  # pylint: disable=duplicate-except
```

Again, it's an edge-case but there's a chance that a delete request was in flight and completed between the Get and Delete operations.

We see things like this much more with S3 based services, so while I generally wouldn't try to catch this edge case, because this is S3, I would.

plugins/modules/s3_metrics_configuration.py
```diff
@@ -0,0 +1,246 @@
...
+    bucket_name = module.params.get('bucket_name')
+    mc_id = module.params.get('id')
+    filter_prefix = module.params.get('filter_prefix')
+    filter_tags = module.params.get('filter_tags')
+
+    try:
+        metrics_configuration = client.get_bucket_metrics_configuration(Bucket=bucket_name, Id=mc_id)
+    except is_boto3_error_code('NoSuchConfiguration'):
+        metrics_configuration = None
+    except ClientError as e:  # pylint: disable=duplicate-except
```
tremble, line 167:
```suggestion
    except (BotoCoreError, ClientError) as e:  # pylint: disable=duplicate-except
```

tremble (changes requested):

It looks like there may still be a missing permission (testing locally I found that DeleteBucketMetricsConfiguration was still missing).

I think there's a few places where the logic is currently correct but could be simplified or reordered for clarity, and it's generally better to catch both ClientError and BotoCoreError.

vorotech:
hi @tremble thanks for looking and providing your feedback.

> It looks like there may still be a missing permission (testing locally I found that DeleteBucketMetricsConfiguration was still missing).

I'm not sure about this. The [documentation says](https://docs.aws.amazon.com/cli/latest/reference/s3api/delete-bucket-metrics-configuration.html) to delete bucket metrics configuration must have permissions to perform the `s3:PutMetricsConfiguration` action.

<details>
<summary>IAM Policy S3 Actions screen</summary><p>

There is no Delete action available.

image
</p></details>

plugins/modules/s3_metrics_configuration.py
```diff
@@ -0,0 +1,221 @@
...
+        module.fail_json_aws(e, msg="Failed to get bucket metrics configuration")
+
+    new_configuration = _create_metrics_configuration(mc_id, filter_prefix, filter_tags)
+
+    if metrics_configuration:
+        if metrics_configuration == new_configuration:
+            module.exit_json(changed=False)
+        
+        if module.check_mode:
+            module.exit_json(changed=True)
```
tremble, line 155:
```suggestion
    if module.check_mode:
        module.exit_json(changed=True)
```

The current indentation means that even in check_mode, if no configuration exists, then we'll add it.

tremble (changes requested):

plugins/modules/s3_metrics_configuration.py
vorotech: Nice spot 👍

vorotech:
@tremble 
Looks good now. This is my first contribution to ansible aws collections. 
I still learn some things. I'm not sure why some of the tests passed, but other configuration failed with `Unable to locate credentials`.

tremble:
The credentials related failure is likely due to the way Ansible's handling new modules.  @jillr @s-hertel do you know if full support for metadata/runtime.yml made it into Ansible 2.9?  Not supporting module_default groups is going to get really annoying with new modules for the next 6 months or so...

vorotech:
Hey @tremble @jillr @s-hertel 
please let me know if I can implement anything to unblock the module

jillr:
2.9 still used ansible/ansible's lib/ansible/config/module_defaults.yml, so those have to be Pr'ed directly to stable-2.9 over there.  I've opened a PR for that and will re-run these tests once that merges.  Sorry for the oversight, I'm still getting used to that!

vorotech:
hey @jillr 
have updated my branch with the latest changes, there are still failed tests, most of them are not related to this module, but only one which hanged (and was terminated after timeout) on attempt to list s3 bucket configuration metrics and I honestly don't know how to resolve such type of issue
https://app.shippable.com/github/ansible-collections/community.aws/runs/954/18/console

vorotech:
Thanks, I see tests has passed now 👍

jillr (approved):
Thanks @vorotech, this lgtm  - @tremble any comments before I merge this new module?

tremble (approved):

tremble:
I think we're good, thanks for your contribution @vorotech.

PR #235, opened 2020-09-18
Fix timeout and Windows AMI in aws_ssm

jillr:
##### SUMMARY

Windows AMIs get replaced every few months, don't be specific with image name

Task timeout feature in ansible-base (#69284) resulted in timeout key not being
properly passed to plugins. This was fixed in (#71722). The intent for the
timeout option in this plugin is to not use Ansible's default timeout however,
but to defer to the inventory var:
https://github.com/ansible/ansible/pull/49652/commits/44f1427507d588d2748c7db34e590234de661564
Rename this internal value from `timeout` to `plugin_timeout`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
aws_ssm

### DIFF
--- a/changelogs/fragments/234-fix_ssm_inventory_plugin_timeout_var.yaml
+++ b/changelogs/fragments/234-fix_ssm_inventory_plugin_timeout_var.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - aws_ssm connection plugin - Change the (internal) variable name from timeout to plugin_timeout to avoid conflicts with ansible/ansible default timeout (#69284, #71722). Developers subclassing this plugin will need to update accordingly.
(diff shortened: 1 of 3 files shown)


jillr:
@psharkey Am I understanding the commit message in https://github.com/ansible/ansible/pull/49652/commits/44f1427507d588d2748c7db34e590234de661564 correctly that the SSM plugin does not intend users to use the `--timeout` cli option?  The change I have here breaks that, and due to changes in ansible/ansible we either have to defer to that value or make our option be entirely separate.  I believe what I have here preserves the intended or more desirable behaviour.

jillr:
Integration tests for the aws_ssm plugin are disabled in CI - they pass for me locally.

psharkey:
@jillr Yes, this was intended to be separate from the `--timeout` cli option and what you have here should preserve the desired behaviour.

PR #271, opened 2020-10-20
Put an upper limit on boto3 in CI

jillr:
##### SUMMARY
Ansible 2.9 has a test constraint for openshift >= 0.6.2, < 0.9.0.  We're only specifying a minimum boto3 today and 1.16.x has released, which pulls in a newer urllib3 than that openshift supports.

Example: https://app.shippable.com/github/ansible-collections/community.aws/runs/907/19/console

For now it can be fixed by capping what boto3 we use in CI, though if 2.9 will have extended support we'll need to deal with it eventually.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests/integration/constraints.txt


tremble (approved):
LGTM just waiting on the tests passing

PR #342, opened 2020-12-25
Allow dependabot to check python dependencies

jlosito:
##### SUMMARY

This change allows dependabot to check python dependencies and submit automatic PRs with version bumps in order to keep packages up-to-date.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates

##### ISSUE TYPE

- Chore Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"


jillr:
Hi @jlosito, thanks very much for this patch. For our collections we do not automatically version our dependencies as we need to ensure that the versions of different libraries being specified are tested and functioning to work in a variety of platforms and environments.  When we do make changes to these requirements they need to be documented and communicated to our users as this may require additional porting work on their part.

PR #352, opened 2021-01-09
aws_ssm signed url using v2 and thus aws_ssm generates incompatible curl request to download s3 object for ansible python

ramvalleru:
##### SUMMARY

Issue is with botocore which still uses V2 signature by default instead of V4 signature. This result in malformed CURL http url to download Ansible python script from S3. Issue is documented in botocore https://github.com/boto/botocore/issues/2109.

Include fix to #351

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

aws_ssm connection
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/connection/aws_ssm.py
+++ b/plugins/connection/aws_ssm.py
@@ -521,6 +521,7 @@ class Connection(ConnectionBase):
             aws_session_token=aws_session_token,
             region_name=region_name,
             config=Config(signature_version="s3v4")
+        )
         return client
 
     @_ssm_retry
--- a/plugins/connection/aws_ssm.py
+++ b/plugins/connection/aws_ssm.py
@@ -162,7 +162,6 @@ EXAMPLES = r'''
 import os
 import getpass
 import json
-import os
 import pty
 import random
 import re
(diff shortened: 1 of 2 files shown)


plugins/connection/aws_ssm.py
```diff
@@ -497,7 +497,11 @@ def _flush_stderr(self, subprocess):
 
     def _get_url(self, client_method, bucket_name, out_path, http_method):
         ''' Generate URL for get_object / put_object '''
-        client = self._get_boto_client('s3')
+        if self.get_option('region') is None:
+            region_name = 'us-east-1'
+        else:
+            region_name = self.get_option('region')
```
goneri, line 503: You can actually just do, and this will also cover the case where the `region` is an empty string:

```python
region_name = self.get_option('region') or 'us-east-1'
```

goneri (changes requested):
Thank you for the Pull Request. We need a changelog fragment before we can merge it. Can you please add one? https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs

plugins/connection/aws_ssm.py
ramvalleru: This has been addressed with the new commits.

ramvalleru:
> Thank you for the Pull Request. We need a changelog fragment before we can merge it. Can you please add one? https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs

Change log set has been updated and pushed.

goneri (approved):

goneri:
Thank you @ramvalleru for your contribution :-).

PR #375, opened 2021-01-27
Add tag query to lambda_info module

ngoduykhanh:
##### SUMMARY

This PR to add tags query into the `lambda_info` module. It is helpful to be able to query the Lambda function's tag list so we can integrate future actions.
This PR also fix some missing parameter

##### ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME

- lambda_info

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/docs/community.aws.lambda_info_module.rst
+++ b/docs/community.aws.lambda_info_module.rst
@@ -222,6 +222,36 @@ Parameters
             <tr>
                 <td colspan="1">
                     <div class="ansibleOptionAnchor" id="parameter-"></div>
+                    <b>max_items</b>
+                    <a class="ansibleOptionLink" href="#parameter-" title="Permalink to this option"></a>
+                    <div style="font-size: small">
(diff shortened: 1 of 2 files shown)


tremble (changes requested):
Hi @ngoduykhanh

Thanks for taking the time to raise this PR.  I apologise for the delay in getting this reviewed.

A few things:
1. Rather than adding the `max_items` and `next_marker` parameters it's generally better to use the boto3 paginators (see https://github.com/ansible-collections/community.aws/pull/442/files for an example)
2. Please add a changelog fragment: https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs-how-to
3. Please update the integration tests to include some examples of setting/fetching tags (tests/integration/targets/lambda/tasks/main.yml)

There's also no need to update docs/community.aws.lambda_info_module.rst this will get updated automatically when we push the next release.

PR #378, opened 2021-01-30
[AWS Cloudformation Lookup] Added aws_cloudformation lookup for retri…

ethemcemozkan:
##### SUMMARY
This PR creates a new lookup named aws_cloudformation. I happen to look up and use AWS Cloudformation stack outputs a lot in my ansible playbook and I thought it would be really useful if we have a quick lookup plugin for this. I think this might be quicker and cleaner instead of retrieving the stack with the cloudformation_info module.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
aws_cloudformation

##### ADDITIONAL INFORMATION
I am not sure about the plugin name. If you have other ideas please let me know and I can refactor the code.


pabelanger:
recheck

gravesm:
@ethemcemozkan thank you for the PR. Could you elaborate on what the benefits are of this over the info module?

ethemcemozkan:
> @ethemcemozkan thank you for the PR. Could you elaborate on what the benefits are of this over the info module?

@gravesm well I always liked the simplicity of the lookup plugins. It keeps your playbook simple when you just need value of the one output.

Also current info module returns outputs in a dict with stack name as key. Which in my opinion creates long unnecessary code like
`register.cloudformation[stack_name].stack_outputs `
We are already specifying stack name in the cloudformation info task. Using stack name again in the `cloudformation[stack_name] ` part looks unnecessary.

PR #386, opened 2021-02-02
ec2_vpc_igw: increase integration tests coverage and code cleaning

alinabuzachis:
##### SUMMARY
 Increase integration test coverage for `ec2_vpc_igw` and `ec2_vpc_igw_info` and some code cleaning

 * Add CamelCase tags integration test for `ec2_vpc_igw` 
 * Removed check_input_tags() because useless as ansible_dict_to_boto3_tag_list() already performs a to_native() conversion to strings
 * Add additional integration tests for `ec2_vpc_igw_info` module

##### COMPONENT NAME
ec2_vpc_igw
ec2_vpc_igw_info

### DIFF
--- a/tests/integration/targets/ec2_vpc_igw/tasks/main.yml
+++ b/tests/integration/targets/ec2_vpc_igw/tasks/main.yml
@@ -37,6 +37,13 @@
     assert:
       that:
           - vpc_result is successful
+          - '"vpc" in vpc_result'
+          - '"id" in vpc_result.vpc'
+          - vpc_result.vpc.state == 'available'
+          - '"tags" in vpc_result.vpc'
+          - vpc_result.vpc.tags | length == 2
+          - vpc_result.vpc.tags["Name"] == "{{ resource_prefix }}-vpc"
+          - vpc_result.vpc.tags["Description"] == "Created by ansible-test"
 
   # ============================================================
   - name: Search for internet gateway by VPC - no matches
@@ -50,6 +57,7 @@
       that:
           - igw_info is successful
           - '"internet_gateways" in igw_info'
+          - (igw_info.internet_gateways | length) == 0
 
   # ============================================================
   - name: create internet gateway (expected changed=true) - CHECK_MODE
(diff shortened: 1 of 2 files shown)


tests/integration/targets/ec2_vpc_igw/tasks/main.yml
```diff
@@ -321,6 +321,109 @@
...
+    ec2_vpc_igw_info:
+      filters:
+        "tag:Title Case": 'Hello Cruel World ❤️'
+    register: igw_info
+
+  - name: Assert success
+    assert:
+      that:
+          - igw_info is successful
+          - '"internet_gateways" in igw_info'
```
jillr, line 383: It would be good to also asset that the gateway that was returned is the ID that we expected whenever we query, just to be safe.
Something like: `igw_info.internet_gateways.internet_gateway_id == igw_id`
If we only filter on tags and more than one instance of the test is running at the same time I think it's possible we could get more than one result.  If that happens you might need to use a tag that has the resource_prefix.

alinabuzachis: Thank you for reviewing @jillr. I applied this change.

tests/integration/targets/ec2_vpc_igw/tasks/main.yml
```diff
@@ -321,6 +321,107 @@
...
+          - igw_info.internet_gateways | length == 0
+
+  - name: Gather information about a filtered list of Internet Gateways using tags (no match) - CHECK_MODE
+    ec2_vpc_igw_info:
```
jillr, line 413: Also need to specify check_mode here

tests/integration/targets/ec2_vpc_igw/tasks/main.yml
```diff
@@ -321,6 +321,107 @@
...
+
+  - name: Assert success
+    assert:
+      that:
+          - igw_info is successful
+          - '"internet_gateways" in igw_info'
+          - igw_info.internet_gateways | selectattr("internet_gateway_id",'equalto',"{{ igw_id }}")
+  
+  - name: Gather information about a filtered list of Internet Gateways using tags - CHECK_MODE
+    ec2_vpc_igw_info:
```
jillr, line 386: You don't actually have check_mode on this task

jillr (changes requested):
Two small notes.

Let's also add an assertion to `Search for internet gateway by VPC - no matches` (the line 42 test task) that igw_info.internet_gateways has a length of zero, to confirm there really are no matches.  Thanks @alinabuzachis!

tests/integration/targets/ec2_vpc_igw/tasks/main.yml
alinabuzachis: @jillr thank you for reviewing. I added the check_mode.

tests/integration/targets/ec2_vpc_igw/tasks/main.yml
alinabuzachis: @jillr thank you for reviewing. I added the check_mode.

tremble (approved):
LGTM and @jillr's review comments have been addressed

PR #457, opened 2021-03-04
add coverage as a test requirement

goneri:
Add the `coverage` module in the test-requirements.txt file. The zuul unit-tests job depends on it.


goneri:
recheck

PR #472, opened 2021-03-12
ec2_vpc_nat_gateway_info: module stabilization 

alinabuzachis:
##### SUMMARY
ec2_vpc_nat_gateway_info module stabilization and cleanup
* Catch and handle (botocore.exceptions.BotoCoreError, botocore.exceptions.ClientError) on boto API calls
* Add paginator
* Document returned data

##### ADDITIONAL INFORMATION

ec2_vpc_nat_gateway_info

### DIFF
--- a/plugins/modules/ec2_vpc_nat_gateway_info.py
+++ b/plugins/modules/ec2_vpc_nat_gateway_info.py
@@ -167,7 +167,7 @@ def _describe_nat_gateways(client, module, **params):
     except is_boto3_error_code('InvalidNatGatewayID.NotFound'):
         module.exit_json(msg="NAT gateway not found.")
     except is_boto3_error_code('NatGatewayMalformed'):  # pylint: disable=duplicate-except
-        module.exit_json(msg="NAT gateway id is malformed.")
+        module.fail_json_aws(msg="NAT gateway id is malformed.")
 
 
 def get_nat_gateways(client, module):
--- a/plugins/modules/ec2_vpc_nat_gateway_info.py
+++ b/plugins/modules/ec2_vpc_nat_gateway_info.py
@@ -30,7 +30,6 @@ author: Karen Cheng (@Etherdaemon)
 extends_documentation_fragment:
 - amazon.aws.aws
 - amazon.aws.ec2
-
 '''
 
 EXAMPLES = r'''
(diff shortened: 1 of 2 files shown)


plugins/modules/ec2_vpc_nat_gateway_info.py
```diff
@@ -69,51 +68,126 @@
...
-def date_handler(obj):
-    return obj.isoformat() if hasattr(obj, 'isoformat') else obj
+@AWSRetry.jittered_backoff(retries=10)
+def _describe_nat_gateways(client, **params):
+    try:
+        paginator = client.get_paginator('describe_nat_gateways')
+        return paginator.paginate(**params).build_full_result()['NatGateways']
+    except is_boto3_error_code('InvalidNatGatewayID.NotFound'):
+        return list()
+    except is_boto3_error_code('NatGatewayMalformed'):  # pylint: disable=duplicate-except
```
jillr, line 169: In this case we should exit and send the error message to the user instead of an empty list; maybe they typoed the gateway ID in their playbook and it could be hard to troubleshoot what's happening if we don't tell them this is a malformed ID.

plugins/modules/ec2_vpc_nat_gateway_info.py
```diff
@@ -122,7 +196,7 @@ def get_nat_gateways(client, module, nat_gateway_id=None):
 def main():
     argument_spec = dict(
         filters=dict(default={}, type='dict'),
-        nat_gateway_ids=dict(default=[], type='list', elements='str'),
+        nat_gateway_ids=dict(default=[], type='list', elements='str')
```
jillr, line 199: Please keep the trailing comma here.  By having a comma already on the last item in this dict, the next time someone adds an option to the module they won't have to add it to this line which reduces the number of lines in the diff when someone reviews it.

jillr (commented):
2 small notes otherwise this looks really good. Thanks @alinabuzachis!

plugins/modules/ec2_vpc_nat_gateway_info.py
alinabuzachis: @jillr Thank you for reviewing. I added the trailing comma.

plugins/modules/ec2_vpc_nat_gateway_info.py
alinabuzachis: @jillr I added added exit_json with the error message. Thank you.

jillr: This should be module.fail_json_aws, so that a task with a malformed ID will fail and the user's playbook would not continue.  Passing a malformed gateway ID would be an error.

alinabuzachis: @jillr, you're right. Just changed. Thank you.

jillr (approved):

PR #473, opened 2021-03-13
Stabilize and improve ec2_vpc_endpoint modules

jillr:
##### SUMMARY
- Add tagging support
- Make idempotent
- Better exception handling
- Better check_mode support
- Use module_utils for common functions
- Enable retries on common AWS failures

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
ec2_vpc_endpoint
ec2_vpc_endpoint_info

### DIFF
--- a/plugins/modules/ec2_vpc_endpoint.py
+++ b/plugins/modules/ec2_vpc_endpoint.py
@@ -66,6 +66,19 @@ options:
     default: present
     choices: [ "present", "absent" ]
     type: str
+  tags:
+    description:
+      - A dict of tags to apply to the internet gateway.
+      - To remove all tags set I(tags={}) and I(purge_tags=true).
+    type: dict
+    version_added: 1.5.0
+  purge_tags:
+    description:
+      - Delete any tags not specified in the task that are on the instance.
+        This means you have to specify all the desired tags on each task affecting an instance.
+    default: false
+    type: bool
+    version_added: 1.5.0
   wait:
     description:
       - When specified, will wait for either available status for state present.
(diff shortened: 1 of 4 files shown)


plugins/modules/ec2_vpc_endpoint.py
```diff
@@ -200,58 +213,111 @@
...
-        try:
-            resource = get_endpoints(client, module, resource_id)['VpcEndpoints'][0]
-            if resource['State'] == status:
-                status_achieved = True
-                break
-            else:
-                time.sleep(polling_increment_secs)
-        except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
-            module.fail_json_aws(e, msg='Failure while waiting for status')
+def match_endpoints(module, endpoint):
```
goneri, line 248: You use module just to get access to the `route_table_ids` and `service` parameters. In order to reduce the inter-dependencies, I suggest to directly pass this two values to the function.
By doing so, it would be easier to write an unit-test.

alinabuzachis (commented):
@jillr It looks good to me.

goneri (approved):

plugins/modules/ec2_vpc_endpoint.py
goneri: To be clear, this is a nitpicking level suggestion as I know we use to pass `module` everywhere.

jillr: It's a good point though, and easily changed - thanks

tremble (changes requested):

PR #484, opened 2021-03-19
Cleanup ec2_vpc_route_table modules

jillr:
##### SUMMARY

* Move regex definitions from global to calling function
* Remove unused regex
* Tidy up comments
* Replace json_query uses in tests with jinja selectattr (See: https://github.com/ansible-collections/amazon.aws/issues/289)
* Additional assertions in integration tests

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ec2_vpc_route_table

### DIFF
--- a/plugins/modules/ec2_vpc_route_table.py
+++ b/plugins/modules/ec2_vpc_route_table.py
@@ -245,11 +245,6 @@ from ansible_collections.amazon.aws.plugins.module_utils.ec2 import compare_aws_
 from ansible_collections.amazon.aws.plugins.module_utils.waiters import get_waiter
 
 
-CIDR_RE = re.compile(r'^(\d{1,3}\.){3}\d{1,3}/\d{1,2}$')
-SUBNET_RE = re.compile(r'^subnet-[A-z0-9]+$')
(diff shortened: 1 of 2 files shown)


plugins/modules/ec2_vpc_route_table.py
```diff
@@ -283,10 +278,10 @@ def find_subnets(connection, module, vpc_id, identified_subnets):
     """
     Finds a list of subnets, each identified either by a raw ID, a unique
     'Name' tag, or a CIDR such as 10.0.0.0/8.
-
-    Note that this function is duplicated in other ec2 modules, and should
-    potentially be moved into a shared module_utils
     """
+    CIDR_RE = re.compile(r'^(\d{1,3}\.){3}\d{1,3}/\d{1,2}$')
+    SUBNET_RE = re.compile(r'^subnet-[A-z0-9]+$')
+
```
goneri, line 284: Yes another candidate for Py3's `ipaddress` module :-).

goneri (approved):

plugins/modules/ec2_vpc_route_table.py
jillr: Yeah I went looking for something existing we could reuse, but as far as I found nothing is there today

PR #525, opened 2021-04-05
Fix route53 idempotency issues

### LINKED ISSUE #524: route53 idempotency issues
##### SUMMARY
Since #405, route53 has several idempotency issues when running with `-C --diff`:
1. it tries to change CAA entries where the values are in a different order;
2. it tries to re-add wildcard records that already exist.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
route53

##### ANSIBLE VERSION
```paste below
devel
```

felixfontein:
##### SUMMARY
Fixes #524.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
route53


tremble (changes requested):
The code change itself looks good.  We just need to get some tests in place.

felixfontein:
The CAA order problem cannot be tested with the module - resp. only parts of it can be tested, and that test is already there (and the module satisfied that test). To fully test this, one needs to create CAA records in **a different** order than the module would create them. Since the module always sorts the CAA values, that's not possible with the module.

tremble (approved):
Thanks for following up on this @felixfontein

felixfontein:
@tremble thanks for reviewing and merging!

PR #565, opened 2021-05-04
Implement tags support for AWS Route 53 objects

badnetmask:
##### SUMMARY
Implement tags support for AWS Route 53 objects.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
community.aws.route53
community.aws.route53_health_check

##### ADDITIONAL INFORMATION
Current AWS Route 53 modules does not support tagging objects. This is important for organization, ownership tracking and financial accounting on enterprise environments.
```
- name: Create zone example.com
  community.aws.route53_zone:
    zone: example.com
    tags:
      Owner: DevOps Team
    purge_tags: True

- name: Create www health check
  community.aws.route53_health_check:
    fqdn: www.example.com
    type: TCP
    port: 443
    tags:
      Owner: DevOps Team
    purge_tags: True
```


badnetmask:
There appears to be one pending error on `ansible-test-cloud-integration-aws-py36`, which I cannot reproduce, but given the message it seems to be related to the test account permissions.

```
raise error_class(parsed_response, operation_name)\nbotocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the ListTagsForResource operation: User: arn:aws:sts::966509639900:assumed-role/ansible-core-ci-test-dev/dev=remote=goneri is not authorized to perform: route53:ListTagsForResource on resource: arn:aws:route53:::hostedzone/Z03802332J78O2JWLQPLE\n",
```

jillr:
recheck

plugins/modules/route53_health_check.py
```diff
@@ -77,6 +77,16 @@
...
+  tags:
+    description:
+      - A hash/dictionary of tags to add to the new instance or to add/remove from an existing one.
+    type: dict
+  purge_tags:
+    description:
+      - Delete any tags not specified in the task that are on the health check.
+        This means you have to specify all the desired tags on each task affecting a health check.
+    default: false
+    type: bool
```
jillr, line 89: We should also note the boto3 requirement for tags in the module docs

plugins/modules/route53_health_check.py
```diff
@@ -121,12 +131,16 @@
...
+    import botocore
 except ImportError:
-    pass  # Handled by HAS_BOTO
+    pass  # Handled by ec2.HAS_BOTO and ec2.HAS_BOTO3
 
 from ansible_collections.amazon.aws.plugins.module_utils.core import AnsibleAWSModule
 from ansible_collections.amazon.aws.plugins.module_utils.ec2 import HAS_BOTO
+from ansible_collections.amazon.aws.plugins.module_utils.ec2 import HAS_BOTO3
 from ansible_collections.amazon.aws.plugins.module_utils.ec2 import get_aws_connection_info
+from ansible_collections.community.aws.plugins.module_utils.route53_tags import manage_tags
```
jillr, line 143:
```suggestion
from ansible_collections.community.aws.plugins.module_utils.route53 import manage_tags
```
I'd rather we name the module_utils file as just `route53.py` so it can readily/obviously be used for other utilities if any are created in the future.

plugins/modules/route53_zone.py
```diff
@@ -48,6 +47,16 @@
...
+    tags:
+        description:
+            - A hash/dictionary of tags to add to the new instance or to add/remove from an existing one.
+        type: dict
+    purge_tags:
+        description:
+            - Delete any tags not specified in the task that are on the zone.
+              This means you have to specify all the desired tags on each task affecting a zone.
+        default: false
+        type: bool
```
jillr, line 59:
```suggestion
        type: bool
        version_added: 2.0.0
```

plugins/modules/route53_zone.py
```diff
@@ -48,6 +47,16 @@
...
 extends_documentation_fragment:
 - amazon.aws.aws
 - amazon.aws.ec2
```
jillr, line 62: Could you please add some EXAMPLES that use `tags` and `purge_tags` in the next section, and update the RETURN block for tags?

plugins/modules/route53_zone.py
```diff
@@ -48,6 +47,16 @@
             - The reusable delegation set ID to be associated with the zone.
             - Note that you can't associate a reusable delegation set with a private hosted zone.
         type: str
+    tags:
+        description:
+            - A hash/dictionary of tags to add to the new instance or to add/remove from an existing one.
+        type: dict
```
jillr, line 53:
```suggestion
        type: dict
        version_added: 2.0.0
```

jillr (changes requested):
This change looks pretty good, thanks @badnetmask.  It would be nice to have some additional tests that change the tags for a resource and that test purge_tags with both true and false.

tests/integration/targets/route53_zone/tasks/main.yml
```diff
@@ -42,6 +45,8 @@
         zone: "{{ resource_prefix }}.check.public"
         comment: original comment
         state: present
+        tags:
+          TestTag: "{{ resource_prefix }}"
```
markuman, line 49: can you also replace `TestTag` with a different Keyname using `purge_tags: yes`?  
Otherwise `purge_tags` is only tested like "remove all keys".

markuman (changes requested):
LGTM. I'd like only to see an addition `purge_tag` integration test, where existing tags will be replaced by other tags.

tests/integration/targets/route53_zone/tasks/main.yml
badnetmask: Done.

markuman (approved):
LGTM

goneri:
recheck

tremble (changes requested):
Hi @badnetmask sorry for the delays here.

With route53_health_check being boto2 based I'm nervous to add changes, especially without integration tests.  

I'd also like to see a couple of non-camel-case tagging examples (at least a "snake_case" example).   There's some easy mistakes that can be made but hidden.  If you remove the route53_health_check change and add a couple of additional tagging tests we can get this merged.

For route53_health_check I've got initial integration tests in #732 to support a migration to boto3 with #734 after which your route53 helper should be really easy to add tagging support into route53_health_check.

alinabuzachis:
recheck

PR #603, opened 2021-06-16
ssm connection plugin - allow s3 bucket to use it's own region setting

rrusso1982:
##### SUMMARY
When using ssm to connect to systems in aws it is required that we utilize an s3 bucket to transfer files from the source to the destination server.  When the bucket resides in a different region than the destination server the wrong s3 endpoint is being selected.  This adds a new configuration value, ansible_aws_ssm_bucket_region, to allow the s3 buckets region to be set directly allowing the transfer to occur as would be expected.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/connection/aws_ssm.py

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,12 @@ community.aws Release Notes
 
 .. contents:: Topics
 
+v2.0.0
+======
+
+Minor Changes
+-------------
+- aws_ssm connection plugin - added support for an independent bucket region (https://github.com/ansible-collections/community.aws/pull/603).
 
 v1.5.0
 ======
--- a/plugins/connection/aws_ssm.py
+++ b/plugins/connection/aws_ssm.py
@@ -513,11 +513,10 @@ class Connection(ConnectionBase):
 
     def _get_url(self, client_method, bucket_name, out_path, http_method, profile_name):
         ''' Generate URL for get_object / put_object '''
-        if self.has_option('bucket_region'):
-            region_name = self.get_option('bucket_region')
-        elif self.has_option('region'):
+        region_name = self.get_option('bucket_region')
+        if region_name is None:
             region_name = self.get_option('region')
-        else:
+        if region_name is None:
             region_name = 'us-east-1'
 
         client = self._get_boto_client('s3', region_name=region_name, profile_name=profile_name)
--- a/plugins/connection/aws_ssm.py
+++ b/plugins/connection/aws_ssm.py
@@ -52,6 +52,7 @@ options:
     description: The region of the S3 bucket used for file transfers.
     vars:
     - name: ansible_aws_ssm_bucket_region
+    version_added: 2.0.0
   plugin:
     description: This defines the location of the session-manager-plugin binary.
     vars:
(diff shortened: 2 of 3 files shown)


wilinger:
Would also be nice for s3 bucket to have its own aws_profile that's different than the var:ansible_aws_ssm_profile for the connection so you don't have to create a new bucket if you have multiple AWS accounts. Similar to aws_secrets plugin.

plugins/connection/aws_ssm.py
```diff
@@ -48,6 +48,10 @@
     description: The name of the S3 bucket used for file transfers.
     vars:
     - name: ansible_aws_ssm_bucket_name
+  bucket_region:
+    description: The region of the S3 bucket used for file transfers.
+    vars:
+    - name: ansible_aws_ssm_bucket_region
```
alinabuzachis, line 54: Could you add please `version_added: 2.0.0`?

alinabuzachis (changes requested):
@rrusso1982 Thank you for your contribution. Could you please add a changelog https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/changelog.html?

plugins/connection/aws_ssm.py
rrusso1982: all set.  Thank you

goneri:
recheck

CHANGELOG.rst
```diff
@@ -4,6 +4,12 @@ community.aws Release Notes
 
 .. contents:: Topics
 
+v2.0.0
```
markuman, line 7: I think the CHANGELOG.rst is somewhat autogenerated. You need to revert this and add a changelogs/fragments file

markuman (changes requested):
@rrusso1982 Can you expand the integration test `tests/integration/targets/aws_ssm_parameter_store/tasks/main.yml` and replace one of the `region=ec2_region` with `bucket_region=ec2_region` to see that CI is still passing?  
So that CI is taking care about this parameter in further changes. Otherwise it looks good to me.

gillg:
Hello !

What is blocking this PR ?
It seems very important to make ssm plugin usable with an inventory accross different regions...
We can't define one bucket by region (and it's a little bit complex...) so it's important to fix the bucket region to avoid this kind of error in curl -GET "https://bucket/script.py" :
```xml
<Code>AuthorizationQueryParametersError</Code>
<Message>Error parsing the X-Amz-Credential parameter; the region 'us-west-2' is wrong; expecting 'us-east-1'</Message>
```
In the case where inside a dynamic inventpry you use : 
```yaml
plugin: amazon.aws.aws_ec2
[...]
compose:
  ansible_aws_ssm_region: placement.region
```

markuman:
@gillg the lack of integration tests and files are touched that must not be touched in feature/bug requests.  
Do you like to adopt and complete this PR?

gillg:
duplicate, can be close

alinabuzachis:
Closed by duplicate https://github.com/ansible-collections/community.aws/pull/854

PR #619, opened 2021-07-01
cdaws_ssm - rename retries setting

samdoran:
##### SUMMARY

There was a bug in Ansible that would result in task retries overriding the connection retries. Using a different name for the connection retries setting avoids this issue on versions of Ansible that are missing the bug fix.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`plugins/connection/aws_ssm.py`

##### ADDITIONAL INFORMATION
Related to https://github.com/ansible/ansible/pull/75155.


goneri:
recheck

jillr (approved):
Thanks @samdoran!

PR #622, opened 2021-07-03
test: aws_msk_cluster is slow

goneri:
The aws_msk_cluster test takes a bit of time. Let's be sure this does
not become a problem in the future by isolating it on its own
dedicated job.


goneri:
recheck

PR #697, opened 2021-08-24
iam_role: replace UpdateRoleDescription with use UpdateRole

mandar242:
##### SUMMARY

Modified `iam_role.py` to replace `update_role_description` (deprecated) with `update_role` action call.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

iam_role

### DIFF
--- a/changelogs/fragments/697-iam_role-replace-UpdateRoleDescription-with-UpdateRole.yml
+++ b/changelogs/fragments/697-iam_role-replace-UpdateRoleDescription-with-UpdateRole.yml
@@ -1,2 +1,2 @@
 bugfixes:
-- iam_role - Modified iam_role to replace update_role_description with update_role (https://github.com/ansible-collections/community.aws/pull/697).
+- iam_role - Modified iam_role internal code to replace update_role_description with update_role (https://github.com/ansible-collections/community.aws/pull/697).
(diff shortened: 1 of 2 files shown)


tremble (approved):
Change looks good.  update_role looks like it was added in 1.9.19 (well within the collection's 'supported' range but a couple of years after the role was originally added).

I've not run the test suite.  To any other reviewers: please run the test suite prior to gating.

changelogs/fragments/697-iam_role-replace-UpdateRoleDescription-with-UpdateRole.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+- iam_role - Modified iam_role to replace update_role_description with update_role (https://github.com/ansible-collections/community.aws/pull/697).
```
jillr, line 2: Nit, I would call out that this is an internal code change.  Users may not be familiar with the boto functions that we use so they might assume this could affect them.  Something like, 'Modified iam_role internal code to replace...'/

jillr (commented):
Tested locally with a non-CI account (since we dom't allow IAM tests in CI).  Passes and lgtm with the minor review request.

jillr (approved):

PR #699, opened 2021-08-26
Support tagging of IAM users

szb640:
##### SUMMARY
I've added support for tagging IAM users in `iam_user`, similarly to how roles are tagged. Retrieving tags via `iam_user_info` is also supported.

### DIFF
--- a/plugins/modules/iam_user.py
+++ b/plugins/modules/iam_user.py
@@ -46,13 +46,13 @@ options:
       - Tag dict to apply to the user.
     required: false
     type: dict
-    version_added: 2.0.0
+    version_added: 2.1.0
   purge_tags:
     description:
       - Remove tags not listed in I(tags) when tags is specified.
     default: true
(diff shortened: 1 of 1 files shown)


plugins/modules/iam_user.py
```diff
@@ -41,6 +41,16 @@
     default: false
     type: bool
     aliases: ['purge_policy', 'purge_managed_policies']
+  tags:
```
markuman, line 44: version_added is missing. The next will be 2.0.0 I think.

markuman (changes requested):

plugins/modules/iam_user.py
markuman: for `tags` and also for `purge_tags`

szb640: I've added `version_added: 2.0.0` to both tags.

PR #775, opened 2021-10-21
fix changelog fragment location

markuman:
`git mv changelogs/767-iam_role-add-waiter-support.yml changelogs/fragments/`


markuman:
> apologies that's my bad !

no worries, errors happen ;)

PR #815, opened 2021-11-30
ec2_asg: enable integration tests, change instance type to a type allowed by policy

mandar242:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1247

##### SUMMARY

Enable ec2_asg integration tests as fix for resolving ec2_asg failing due to policies missing is merged.
Also change the instance type used in couple of assertions to a type which is allowed by the policy to fix failure.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ec2_asg


jillr:
recheck

goneri:
recheck

jillr:
@mandar242 Ah, there's a bug in the tests that only happens when the resource_prefix (which is based on the test system hostname) has dashes at the end.  
```
    - name: load balancer name has to be less than 32 characters
      # the 8 digit identifier at the end of resource_prefix helps determine during which test something
      # was created
      set_fact:
        load_balancer_name: "{{ item }}-lb"
      loop: "{{ resource_prefix | regex_findall('.{8}$') }}"
```
This regex takes whatever the last 8 characters are, without consideration that `LoadBalancer name cannot begin or end with a dash`.

The resource_prefix in CI looks like: `ansible-test-54272280-ip-172-16-141-188` so we end up with a `load_balancer_name` of `-141-188` which gives the ValidationError.  We didn't see this when we tested locally because both of our laptops have hostnames without dashes in them.  ansible-test now has another variable we can use to solve exactly this kind of problem with AWS character limits.  Could you please remove this `set_fact` statement and instead make the tests use ` '{{ tiny_prefix }}-lb'` for the LB name?

jillr (commented):
Spent some time on the failures for the mixed-instances test but I haven't got it sorted out yet.  I suspect it might have something to do with the way AWSServiceRoleForAutoScaling is being used to access the launch template.  Documenting what I've tried here:
Changing the compute policy to allow the `AllowAsgInstancesInstanceType` Sid to access resource `'*'` gets me to a new, more descriptive error that `An error occurred (AccessDenied) when calling the UpdateAutoScalingGroup operation: You are not authorized to use launch template: lt-0a2196568c24cb25d",`.  I'm not finding a whole lot of documentation for PassRole for ASGs but I took a couple tries at allowing AWSServiceRoleForAutoScaling to be passed to ec2 with no success.  I've reverted the CI staging policies back to main in the interim.

@tremble Do you have any experience working with ASG policies and launch templates?

mandar242:
recheck

mandar242:
recheck

mandar242:
recheck

mandar242:
recheck

jillr (approved):

jillr:
thanks @mandar242! good work on this one!

PR #863, opened 2022-01-13
docs fix. clean up duplicated and unspecific requirements

markuman:
##### SUMMARY
* remove unspecific/duplicated requirements

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

* aws_msk_config
* ecs_taskdefinition

##### ADDITIONAL INFORMATION
I don't see the `json` library is explicit used in `aws_msk_config`. Furthermore it's standard python library.

### DIFF
--- a/plugins/modules/aws_msk_config.py
+++ b/plugins/modules/aws_msk_config.py
@@ -12,8 +12,6 @@ DOCUMENTATION = r"""
 module: aws_msk_config
 short_description: Manage Amazon MSK cluster configurations.
 version_added: "2.0.0"
-requirements:
-    - botocore >= 1.17.48
 description:
     - Create, delete and modify Amazon MSK (Managed Streaming for Apache Kafka) cluster configurations.
 author:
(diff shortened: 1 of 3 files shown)


plugins/modules/aws_msk_config.py
```diff
@@ -14,7 +14,6 @@
 version_added: "2.0.0"
 requirements:
     - botocore >= 1.17.48
```
alinabuzachis, line 16: Not sure about this, but since we are requiring `botocore >= 1.19.0`, I guess we can also remove this (if there is a specific option requiring `botocore < 1.19.0`, not sure how we do handle that)

markuman: > I guess we can also remove this (if there is a specific option requiring botocore < 1.19.0, not sure how we do handle that)

I don't see that the integrationtest installs a specific version. So basically CI must fail if it's requiring a lower botocore version.  
I'm going to remove it also.

alinabuzachis (approved):

marknet15 (approved):

tremble (approved):

PR #878, opened 2022-01-20
Fix IOPs io1 DB instance updates and integration tests also

marknet15:
##### SUMMARY
Primary this PR is to fix updates when updating `iops` or `allocated_storage` on `io1` DB instances when only one param is changing. 

Secondarily this fixes up the tests again and is test against some improvements to the waiter configuration see linked PR.

IOPs error on update attempts if only one param is being updated:
```
  error:
    code: InvalidParameterCombination
    message: You must specify both the storage size and iops when modifying the storage size or iops on a DB instance that has iops.
    type: Sender
  msg: 'Unable to modify DB instance: An error occurred (InvalidParameterCombination) when calling the ModifyDBInstance operation: You must specify both the storage size and iops when modifying the storage size or iops on a DB instance that has iops.'
```

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
rds_instance

##### ADDITIONAL INFORMATION
These tests are very slow and still a little flakey but generally all pass as expected now locally.


tests/integration/targets/rds_instance/inventory
```diff
@@ -1,12 +1,18 @@
...
+# other tests
 states
 modify
-modify_complex
-processor_features
-read_replica
-vpc_security_groups
-restore_instance
 tagging
+replica
```
marknet15, line 15: renamed these to ensure we sit in the resource length limits as there was a load of issues on the `3.0.0` release around these limits being hit leading to these tests being disabled.

marknet15:
recheck

marknet15:
recheck

tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_sgroups.yml
```diff
@@ -28,7 +28,6 @@
           - {"cidr": "10.122.122.128/28", "zone": "{{ aws_region }}a"}
           - {"cidr": "10.122.122.144/28", "zone": "{{ aws_region }}b"}
           - {"cidr": "10.122.122.160/28", "zone": "{{ aws_region }}c"}
-          - {"cidr": "10.122.122.176/28", "zone": "{{ aws_region }}d"}
```
marknet15, line 31: some regions don't have a `d` AZ such as `eu-west-1`

marknet15:
recheck

marknet15:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

markuman (approved):

tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_complex.yml
```diff
@@ -97,17 +98,16 @@
...
 
-  always:
-    - name: Delete the instance
-      rds_instance:
-        id: "{{ instance_id }}"
-        state: absent
-        skip_final_snapshot: True
-        wait: false
-      ignore_errors: yes
+  # always:
```
alinabuzachis, line 106: Why that's commented on?

alinabuzachis (commented):
May I ask you to update the RETURN statement please?
I quickly checked and it seems: `"associated_roles": [],` is not documented, while it seems `kms_key_id`is not returned even if documented in the RETURN statement.
Could you please double check?

tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_complex.yml
marknet15: Ah did not mean to commit that ! It was when I was locally testing

marknet15:
> May I ask you to update the RETURN statement please? I quickly checked and it seems: `"associated_roles": [],` is not documented, while it seems `kms_key_id`is not returned even if documented in the RETURN statement. Could you please double check?

@alinabuzachis so I checked and `"kms_key_id": null` can be seen in output logs although I do think there's any examples of it being set in the integration tests currently. I have though added `associated_roles`

alinabuzachis (approved):

Issue #749, opened 2021-10-09
cloudfront_distribution: s3_origin_access_identity_enabled


nick-zh:
### Summary

[Doc states](https://docs.ansible.com/ansible/latest/collections/community/aws/cloudfront_distribution_module.html#parameter-origins/s3_origin_access_identity_enabled) that `s3_origin_access_identity_enabled` is a bool 
A bool results for me in the following error:
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: Unknown parameter in DistributionConfig.Origins.Items[0]: "S3OriginAccessIdentityEnabled", must be one of: Id, DomainName, OriginPath, CustomHeaders, S3OriginConfig, CustomOriginConfig, ConnectionAttempts, ConnectionTimeout, OriginShield
fatal: [localhost]: FAILED! => {"boto3_version": "1.18.47", "botocore_version": "1.21.47", "changed": false, "msg": "Error creating distribution: Parameter validation failed:\nUnknown parameter in DistributionConfig.Origins.Items[0]: \"S3OriginAccessIdentityEnabled\", must be one of: Id, DomainName, OriginPath, CustomHeaders, S3OriginConfig, CustomOriginConfig, ConnectionAttempts, ConnectionTimeout, OriginShield"}
```
I also tried creating an oai with `community.aws.cloudfront_origin_access_identity` and adding an id, same result.

### Issue Type

Bug Report

### Component Name

cloudfront_distribution

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.10.8
  config file = <myProjectPath>/ansible.cfg
  configured module search path = ['<myHome>/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.9.7 (default, Sep 24 2021, 09:43:00) [GCC 10.3.0]

```

### Collection Versions

```console (paste below)
$ ansible-galaxy collection list
Collection    Version
------------- -------
amazon.aws    1.5.0  
community.aws 1.5.0  

# /usr/lib/python3/dist-packages/ansible_collections
Collection                Version
------------------------- -------
amazon.aws                1.4.0  
ansible.netcommon         1.5.0  
ansible.posix             1.1.1  
ansible.windows           1.4.0  
arista.eos                1.3.0  
awx.awx                   14.1.0 
azure.azcollection        1.4.0  
[...]

PR #881, opened 2022-01-22
Fix cloudfront_distribution s3_origin_access_identity_enabled bug

marknet15:
##### SUMMARY
If `s3_origin_access_identity_enabled` is set to `True` but no `s3_origin_config` then a default origin config is applied however it also picks up `s3_origin_access_identity_enabled` as `S3OriginAccessIdentityEnabled` and passes it to the API request which is not a valid option to be passed and then fails validation.

Fixes: https://github.com/ansible-collections/community.aws/issues/749

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cloudfront_distribution

##### ADDITIONAL INFORMATION
The option mention is not valid for the API request:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudfront.html#CloudFront.Client.create_distribution

### DIFF
--- a/changelogs/fragments/881-cloudfront-bug.yml
+++ b/changelogs/fragments/881-cloudfront-bug.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - cloudfront_distribution - Dont pass ``s3_origin_access_identity_enabled`` to API request (https://github.com/ansible-collections/community.aws/pull/881).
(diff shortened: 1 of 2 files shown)


marknet15:
recheck

marknet15:
recheck

markuman (approved):

alinabuzachis (approved):

PR #894, opened 2022-01-27
add check_mode for elb_application_lb* & refactor integration tests

jatorcasso:
##### SUMMARY
Add check_mode support for elb_application_lb* & refactor integration tests.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
elb_application_lb
elb_application_lb_info


tests/integration/targets/elb_application_lb/tasks/main.yml
```diff
@@ -1,12 +1,1137 @@
...
+                Actions:
+                  - TargetGroupName: "{{ tg_name }}"
+                    Type: forward
+      register: elb
+      check_mode: yes
+
+    - assert:
+        that:
+          - elb is not changed
+          - elb.msg is match('IN CHECK MODE - no changes to make to ELB specified.')
```
alinabuzachis, line 461: @jillr Do we want to keep all this messaging for the elb? I guess this comes from module_utils.

alinabuzachis (commented):
Thanks @jatorcasso. I left some initial comments. I will do a more accurate review.

plugins/modules/elb_application_lb.py
```diff
@@ -514,9 +525,12 @@ def create_or_update_elb(elb_obj):
 
     # Listeners
     listeners_obj = ELBListeners(elb_obj.connection, elb_obj.module, elb_obj.elb['LoadBalancerArn'])
-
     listeners_to_add, listeners_to_modify, listeners_to_delete = listeners_obj.compare_listeners()
 
+    # Exit on check_mode
+    if elb_obj.module.check_mode and (listeners_to_add or listeners_to_modify or listeners_to_delete):
+        elb_obj.module.exit_json(changed=True, msg='Would have updated ELB if not in check mode.')
```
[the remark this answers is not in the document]
jatorcasso: @alinabuzachis I stole that idea from [ec2_vol](https://github.com/ansible-collections/amazon.aws/pull/509/files), but I actually really like the idea of having a message there! Is there a standard way of returning on check_mode?

tests/integration/targets/elb_application_lb/tasks/main.yml
```diff
@@ -1,12 +1,1137 @@
...
+      register: public_subnets
+
+    - ec2_vpc_subnet_info:
+        filters:
+          vpc-id: '{{ vpc_id }}'
+      register: vpc_subnets
+
+    - name: Create list of subnet ids
+      set_fact:
+        public_subnets: '{{ vpc_subnets|community.general.json_query(''subnets[?tags.Public == `True`].id'') }}'
```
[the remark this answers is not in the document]
jatorcasso: done!

plugins/modules/elb_application_lb.py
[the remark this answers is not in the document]
jillr: > Is there a standard way of returning on check_mode?

Nope!  :smile: 
In the majority of cases AWS modules will do one of:
* Return a `msg` like you have here
* Return only a `changed` key (based on if there would have been a change, if not in check mode)
* Return something that resembles the normal RETURN of the module (like Alina has noted)

Some of this is just the legacy of 7 or 8 years of organic changes to the AWS modules.  Some of it is what makes the most sense for the way the module code has been structured (since we also don't have a defined standard that we expect for that either).  The Ansible devel guide doesn't dictate a preference for what check_mode should return, and the user guide doesn't promise anything, so it's collection/developer preference.  I'm personally ok with having check_mode return only a `changed` or having a `msg`.

alinabuzachis:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

jatorcasso:
recheck

tests/integration/targets/elb_application_lb_info/tasks/main.yml
```diff
@@ -1,11 +1,250 @@
...
+
+    - name: Create a target group for testing
+      elb_target_group:
+        name: '{{ tg_name }}'
+        protocol: http
+        port: 80
+        vpc_id: '{{ vpc_id }}'
+        state: present
+      register: tg
+
```
alinabuzachis, line 109: Can you please organise the tests as tremble suggested in this comment https://github.com/ansible-collections/amazon.aws/pull/509#issuecomment-932056400?

alinabuzachis (changes requested):

tests/integration/targets/elb_application_lb_info/tasks/main.yml
jatorcasso: yes - wasnt sure if it needed that depth on the info module but I will update

alinabuzachis: I mean for the elb_application_lb module, not for the info module (Sorry, pinned the comment on the wrong file).

alinabuzachis: So, I actually missed the tests for elb_application_lb.

plugins/modules/elb_application_lb.py
```diff
@@ -466,8 +466,9 @@
 '''
```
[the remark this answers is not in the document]
jatorcasso: updated! fixed incorrect param type as well.

plugins/modules/elb_application_lb.py
```diff
@@ -420,22 +430,47 @@
...
     type: str
-    sample: arn:aws:elasticloadbalancing:ap-southeast-2:0123456789:loadbalancer/app/my-elb/001122334455
+    sample: "arn:aws:elasticloadbalancing:ap-southeast-2:0123456789:loadbalancer/app/my-elb/001122334455"
 load_balancer_name:
     description: The name of the load balancer.
     returned: when state is present
     type: str
-    sample: my-elb
+    sample: "my-elb"
+resource_actions:
```
alinabuzachis, line 439: I guess there is no need to document `resource_actions:`.

alinabuzachis (approved):
Other than that, it looks good to me. Thank you @jatorcasso. Let's wait for some other review from @marknet15 @markuman.

plugins/modules/elb_application_lb.py
jatorcasso: I was just covering every parameter returned.. does this mean I should also remove `changed` and `failed` ?

marknet15: Yep don't think other modules have this added here and I don't think its relevant to the module itself per say.

plugins/modules/elb_application_lb_info.py
```diff
@@ -69,77 +69,154 @@
...
+            sample: [{ "load_balancer_addresses": [], "subnet_id": "subnet-aabbccddff", "zone_name": "ap-southeast-2a" }]
         canonical_hosted_zone_id:
             description: The ID of the Amazon Route 53 hosted zone associated with the load balancer.
             type: str
-            sample: ABCDEF12345678
+            sample: "ABCDEF12345678"
+        changed:
+            description: Whether an ELB was created/updated/deleted
+            type: bool
+            sample: true
```
marknet15, line 92: I don't think this is applicable for a info module as nothing would change

plugins/modules/elb_application_lb_info.py
```diff
@@ -69,77 +69,154 @@
...
+            sample: "arn:aws:elasticloadbalancing:ap-southeast-2:0123456789:loadbalancer/app/my-elb/001122334455"
         load_balancer_name:
             description: The name of the load balancer.
             type: str
-            sample: my-elb
+            sample: "my-elb"
+        routing_http2_enabled:
+            description: Indicates whether HTTP/2 is enabled.
+            type: bool
+            sample: true
```
marknet15, line 172: Same as on the earlier comment from @alinabuzachis probably no point including this here

tests/integration/targets/elb_application_lb/tasks/main.yml
```diff
@@ -1,12 +1,1172 @@
...
+        that:
+          - elb is not changed
+          - elb.security_groups[0] == sec_group2.group_id
+
+    # ------------------------------------------------------------------------------------------
+
+    - name: Ensure elb_application_lb_info supports check_mode
+      elb_application_lb_info:
+      register: elb_info
+      check_mode: yes
```
marknet15, line 1001: Not sure what others think but I didn't think check mode was really applicable for info modules as nothing really changes

alinabuzachis: I know that all the modules should support check_mode. You're right, there is no change. _info modules are a particular case, but don't know about the story behind the check_mode here.

marknet15 (approved):

tests/integration/targets/elb_application_lb/tasks/main.yml
marknet15: Yeah makes sense, guess they are indeed quite a particular case

jatorcasso: Yeah, I was basing that off a conversation with @jillr who basically said it should be there so nothing breaks if the user does --check on an entire playbook

marknet15: Ahhh cool cool 👍🏻

jillr (changes requested):
@jatorcasso Can you please retain the ALB naming in this set of modules and tests?  ELBs are generally understood to mean v1/Classic Load Balancers (covered by amazon.aws.elb_classic_lb, the module formerly known as ec2_elb_lb).  Only Application Load Balancers (elbv2) are covered by these modules.
https://docs.aws.amazon.com/elasticloadbalancing/index.html
The term ALB is a widely-used and understood convention for AWS users https://cloudacademy.com/blog/application-load-balancer-vs-classic-load-balancer/

jatorcasso:
> @jatorcasso Can you please retain the ALB naming in this set of modules and tests? ELBs are generally understood to mean v1/Classic Load Balancers (covered by amazon.aws.elb_classic_lb, the module formerly known as ec2_elb_lb). Only Application Load Balancers (elbv2) are covered by these modules. https://docs.aws.amazon.com/elasticloadbalancing/index.html The term ALB is a widely-used and understood convention for AWS users https://cloudacademy.com/blog/application-load-balancer-vs-classic-load-balancer/

@jillr changed to ALB where it made the most sense to me. Please let me know if I missed any, thanks!

PR #880, opened 2022-01-22
Add dynamodb table class support

marknet15:
##### SUMMARY
Add support for defining a `TableClass` on DynamoDB tables.

`TableClass` was introduced as part of `botocore` version `1.23.18`
https://github.com/boto/botocore/blob/develop/CHANGELOG.rst#12318

Fixes: https://github.com/ansible-collections/community.aws/issues/829

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
dynamodb_table

##### ADDITIONAL INFORMATION
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.create_table
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.update_table

PR #917, opened 2022-02-02
ec2_launch_template: implement missing metadata options

markuman:
##### SUMMARY
Add missing metadata options
* `instance_metadata_tags`
* `http_protocol_ipv6`

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
ec2_launch_template

### DIFF
--- a/plugins/modules/ec2_launch_template.py
+++ b/plugins/modules/ec2_launch_template.py
@@ -535,15 +535,19 @@ def create_or_update(module, template_options):
 
     if lt_data.get('MetadataOptions'):
         if not module.boto3_at_least('1.20.30'):
-            # warning if enabled is requested, because disabled is the API default value.
+            # fail only if enabled is requested
             if lt_data['MetadataOptions'].get('InstanceMetadataTags') == 'enabled':
-                module.warn('Metadata option "instance_metadata_tags" is ignored, because it requires boto3 >= 1.20.30.')
+                module.require_boto3_at_least('1.20.30', reason='to set instance_metadata_tags')
+            # pop if it's not requested to keep backwards compatibility.
+            # otherwise the modules failes because parameters are set due default values
             lt_data['MetadataOptions'].pop('InstanceMetadataTags')
 
         if not module.boto3_at_least('1.18.29'):
-            # warning if enabled is requested, because disabled is the API default value.
+            # fail only if enabled is requested
             if lt_data['MetadataOptions'].get('HttpProtocolIpv6') == 'enabled':
-                module.warn('Metadata option "http_protocol_ipv6" is ignored, because it requires boto3 >= 1.18.29.')
+                module.require_boto3_at_least('1.18.29', reason='to set http_protocol_ipv6')
+            # pop if it's not requested to keep backwards compatibility.
+            # otherwise the modules failes because parameters are set due default values
             lt_data['MetadataOptions'].pop('HttpProtocolIpv6')
 
     if not (template or template_versions):
--- a/plugins/modules/ec2_launch_template.py
+++ b/plugins/modules/ec2_launch_template.py
@@ -526,6 +526,9 @@ def delete_template(module):
 
 
 def create_or_update(module, template_options):
+    if not module.boto3_at_least('1.20.46'):
+        template_options['metadata_options']['options'].pop('instance_metadata_tags')
+
     ec2 = module.client('ec2', retry_decorator=AWSRetry.jittered_backoff(catch_extra_error_codes=['InvalidLaunchTemplateId.NotFound']))
     template, template_versions = existing_templates(module)
     out = {}
--- a/plugins/modules/ec2_launch_template.py
+++ b/plugins/modules/ec2_launch_template.py
@@ -535,8 +535,15 @@ def create_or_update(module, template_options):
 
     if lt_data.get('MetadataOptions'):
         if not module.boto3_at_least('1.20.30'):
+            # warning if enabled is requested, because disabled is the API default value.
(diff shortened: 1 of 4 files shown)


plugins/modules/ec2_launch_template.py
```diff
@@ -671,7 +686,9 @@ def main():
             options=dict(
                 http_endpoint=dict(choices=['enabled', 'disabled'], default='enabled'),
                 http_put_response_hop_limit=dict(type='int', default=1),
-                http_tokens=dict(choices=['optional', 'required'], default='optional')
+                http_tokens=dict(choices=['optional', 'required'], default='optional'),
+                http_protocol_ipv6=dict(choices=['disabled', 'enabled'], default='disabled'),
+                instance_metadata_tags=dict(choices=['disabled', 'enabled'], default='disabled'),
```
markuman, line 691: Hm maybe we must pop `metadata_options.instance_metadata_tags` if current used boto3 version does not support it.  
Otherwise all playbook will fail using 3.1.0 when boto3 is to old.

Are there any examples how we check the boto3 version? @tremble  @alinabuzachis

markuman: Ok, done. But it's failing because of `http_protocol_ipv6`  
No idea when it was introduced

markuman:
* instance_metadata_tags
  * https://awsapichanges.info/archive/changes/022017-ec2.html
* http_protocol_ipv6
  * https://awsapichanges.info/archive/changes/0bc224-ec2.html

plugins/modules/ec2_launch_template.py
```diff
@@ -516,6 +532,13 @@ def create_or_update(module, template_options):
     out = {}
     lt_data = params_to_launch_data(module, dict((k, v) for k, v in module.params.items() if k in template_options))
     lt_data = scrub_none_parameters(lt_data, descend_into_lists=True)
+
+    if lt_data.get('MetadataOptions'):
+        if not module.boto3_at_least('1.20.30'):
+            lt_data['MetadataOptions'].pop('InstanceMetadataTags')
+        if not module.boto3_at_least('1.18.29'):
+            lt_data['MetadataOptions'].pop('HttpProtocolIpv6')
```
marknet15, line 547: So in other modules it seems to be more standard to fail or warn that the required boto version isn't used when passing a particular param. Personally I find an explicit fail or warn better to work with so that it's clear on the what's required to resolve etc

tests/integration/targets/ec2_launch_template/tasks/instance-metadata.yml
```diff
@@ -1,24 +1,77 @@
...
+        state: absent
+      register: del_lt
+      retries: 10
+      until: del_lt is not failed
+      ignore_errors: true
+
+- name: test with boto3 version that supports instance_metadata_tags
+  block:
+    - pip:
+        name: virtualenv
```
marknet15, line 35: So there is a helper to get the newer botocore version in a venv that sets the interpreter var for easy use in the tests for example like done here:

https://github.com/ansible-collections/community.aws/pull/880/files

plugins/modules/ec2_launch_template.py
markuman: I've added and pushed minutes ago :)

>  I find an explicit fail 

I'm not sure if we should fail, because the boto3 requirements for the entire collection is lower.  
Furthermore, when we fail, all playbooks that run on 3.0.1, won't run with 3.1.0 anymore when the boto3 library is not updated - just because of parameters that most likely are not requested.

markuman: Maybe we can fail, when other values as the APIs default values are requested ...

markuman: some more opinions cc @alinabuzachis  @tremble  @jillr

marknet15: Awesome think I hadn't refreshed 😀  

Yeah that makes sense it is a little inconsistent across the modules to be fair as some places do seem to explicitly fail and others warn but yeah definitely better to have something to flag it up

tests/integration/targets/ec2_launch_template/tasks/instance-metadata.yml
markuman: Ah nice. Thx

plugins/modules/ec2_launch_template.py
alinabuzachis: I can suggest something like this: if ignoring the new options won't affect the normal behaviour of the module, I'd probably only warn.

alinabuzachis: Not entirely sure of how handle that and I agree with what @marknet15 said.

However, there is an example https://github.com/ansible-collections/amazon.aws/blob/main/plugins/modules/ec2_instance.py#L989

markuman: Ah, there is a `reason` parameter. Thx

markuman: >  if ignoring the new options won't affect the normal behaviour of the module, I'd probably only warn.

Now it 

* failed if `http_protocol_ipv6: enabled` and `instance_metadata_tags: enabled` are requested but the boto3 version is not sufficient
* passed if  `http_protocol_ipv6: disabled` and `instance_metadata_tags: disabled` are requested or not set (default values) and boto3 version is not sufficient. That covers the backwart compatiblity from community.aws < 3.1.0
* passed if  `http_protocol_ipv6: enabled` and `instance_metadata_tags: enabled` are requested  and boto3 version is sufficient.

All three cases are covered in the integration test

marknet15 (approved):

alinabuzachis (approved):

PR #932, opened 2022-02-09
Always return lists in cloudfront_info where it returns a list already…

stefanhorning:
…when length is > 0. So it should also return same type in other case.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cloudfront_info.py

##### ADDITIONAL INFORMATION
Fixes problems when running against empty lists. Check this old PR for reference https://github.com/ansible-collections/community.aws/pull/293

### DIFF
--- a/plugins/modules/cloudfront_info.py
+++ b/plugins/modules/cloudfront_info.py
@@ -327,7 +327,9 @@ class CloudFrontServiceManager:
 
             if len(origin_access_identity_list['Items']) > 0:
                 return origin_access_identity_list['Items']
-            return {}
+            else:
+                return []
+
         except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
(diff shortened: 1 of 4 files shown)


plugins/modules/cloudfront_info.py
```diff
@@ -327,7 +327,9 @@ def list_origin_access_identities(self):
 
             if len(origin_access_identity_list['Items']) > 0:
```
stefanhorning, line 328: I guess here we could drop the entire if statement as origin_access_identity_list['Items'] will always fall back to an empty list (see line above).

stefanhorning:
still needs some manual tests, as it seems that the test target for cloudfront is disabled (on CI).

stefanhorning:
looks like I can not reproduce the error I tried to fix in #293 so I will rather close this.

Issue #649, opened 2021-07-19
ec2_asg: add detach_instances feature


dstan88:
### Summary

Add 'detach_instances' feature that will not terminate the instance(s) after being replaced in the AutoScalingGroup****

### Issue Type

Feature Idea

### Component Name

ec2_asg

### Additional Information

```    
    # Custom check for detaching the EC2 Instance from Autoscaling Group
    if detach_instances:
        instance_to_detach = replace_instances
        changed = False
        for instance_id in instance_to_detach:
            if instance_id in instances:
                changed = True
                detach_asg_instance(connection, instance_id, group_name, False)
                wait_for_new_inst(connection, group_name, wait_timeout, desired_capacity, 'viable_instances')
                wait_for_elb(connection, group_name)
                wait_for_target_group(connection, group_name)

        return(changed, props)
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/ec2_asg.py`](https://github.com/['ansible-collections/amazon.aws', 'ansible-collections/community.aws', 'ansible-collections/community.vmware']/blob/main/plugins/modules/ec2_asg.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @garethr @jillr @markuman @s-hertel @tremble @wimnat
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

alinabuzachis:
Hi @dstan88, thank you for this feature idea. Would you be willing to work on this?

mandar242:
Closing the issue as #933 is merged and resolves this issue.

PR #933, opened 2022-02-10
ec2_asg: Add functionality to detach specified instances from ASG 

mandar242:
##### SUMMARY

Adds feature to detach specified instances from a AutoScalingGroup rather than terminating them directly.
Detached instances are not terminated and can be managed independently.

Implements #649 
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ec2_asg
##### ADDITIONAL INFORMATION

Makes use of 
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling.html#AutoScaling.Client.detach_instances


plugins/modules/ec2_asg.py
```diff
@@ -1523,6 +1542,37 @@ def replace(connection):
...
+    instances = props['instances']
+
+    # check if provided instance exists in asg
+    for instance_id in detach_instances:
+        if instance_id not in instances:
+            module.fail_json(msg="Provided instance ID: {0} does not belong to the AutoScalingGroup: {1}".format(
+                instance_id, group_name))
+
+    # check if minimum size is greater than what setting decrement_desired_capacity would make it
+    if min_size and min_size > len(instances) - len(detach_instances) and decrement_desired_capacity:
```
alinabuzachis, line 1561: Shouldn't be less than `min_size`?

plugins/modules/ec2_asg.py
```diff
@@ -182,6 +182,19 @@
...
+      - If I(decrement_desired_capacity) flag is not set, new instance(s) are lauched to replace the detached instance(s).
+      - If a Classic Load Balancer attached to the AutoScalingGroup, the instances are also deregistered from the load balancer.
+      - If there are target groups attached to the AutoScalingGroup, the instances are also deregistered from the target groups.
+    type: list
+    elements: str
+  decrement_desired_capacity:
+    description:
+      - Indicates whether the AutoScalingGroup decrements the desired capacity value by the number of instances detached.
+    default: false
+    type: bool
```
alinabuzachis, line 197: `version_added: 3.2.0`

alinabuzachis (changes requested):
@mandar242 Thank you for working on this. You should add some integration tests for these features and a changelog fragment.

plugins/modules/ec2_asg.py
```diff
@@ -1523,6 +1542,37 @@ def replace(connection):
...
+    as_group = describe_autoscaling_groups(connection, group_name)[0]
+    decrement_desired_capacity = module.params.get('decrement_desired_capacity')
+    min_size = module.params.get('min_size')
+    props = get_properties(as_group)
+    instances = props['instances']
+
+    # check if provided instance exists in asg
+    for instance_id in detach_instances:
+        if instance_id not in instances:
+            module.fail_json(msg="Provided instance ID: {0} does not belong to the AutoScalingGroup: {1}".format(
```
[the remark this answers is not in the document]
mandar242: Hi @alinabuzachis, without this check, if the instance is not a part of ASG, the boto3 fails and throws the following error
```
msg: "Failed to describe launch configurations: An error occurred (ValidationError) when calling
 the DetachInstances operation: The instance i-123456789 is not part of Auto Scaling group test-asg-xxx.
    "response_metadata": {
        "http_headers": {
            "connection": "close",
            "content-length": "325",
            "content-type": "text/xml",
            "date": "Tue, 15 Feb 2022 19:40:00 GMT",
            "x-amzn-requestid": "f83259fc-d609-498a-a197-xxxxxxx"
        },
        "http_status_code": 400,
        "request_id": "f83259fc-d609-498a-a197-xxxxxxx",
        "retry_attempts": 0
    }
```
to handle this error and making it more easy to understand for user, I believe this check would be helpful as it throws following error in the same case
```
"msg": "Provided instance ID: i-123456789 does not belong to the AutoScalingGroup: test-asg-xxx"
```
but we can go either way on this behavior as you mentioned
1. Stop the execution and fail
2. Throw a warning for the instance that is not part of ASG but continue detaching other instances

@jillr what do you think would be a better option for handling this?

plugins/modules/ec2_asg.py
mandar242: Hi @alinabuzachis , apologies for having a bit weird ```if``` statement, I'll make it more clear right away.

Also to explain this check, consider following example task from above playbook https://github.com/ansible-collections/community.aws/pull/933#issue-1129356479:
```
 # detach 2 instances from the asg and reduce the desired capacity from 3 to 1
    - name: detach 2 instance from the asg and reduce the desired capacity from 3 to 1
      community.aws.ec2_asg:
        name: test-asg-abc
        launch_config_name: my_new_lc
        health_check_period: 60
        health_check_type: ELB
        min_size: 3
        max_size: 5
        desired_capacity: 3
        region: us-east-2
        decrement_desired_capacity: true
        detach_instances:
          - '{{ instances_detach_replace[0] }}'
          - '{{ instances_detach_replace[1] }}'
```

Here, as the ```decrement_desired_capacity``` is set to ```true```, when these ```2``` instances are detached from the ASG, the `desired_capacity` is also decremented from `3` to `1` which will make `desired_capacity` less than the current set `min_size` in the `ASG`.
But as the `min_size` in the `ASG config` is set to `3` is causes the `boto3` to throw following error
```
"msg": "Failed to describe launch configurations: An error occurred (ValidationError) when calling the 
DetachInstances operation: AutoScalingGroup test-asg-xxx has min-size=3, max-size=5, and desired-size=3.
To detach 2 instances,  please update the AutoScalingGroup sizes appropriately.",
    "response_metadata": {
        "http_headers": {
            "connection": "close",
            "content-length": "409",
            "content-type": "text/xml",
            "date": "Tue, 15 Feb 2022 20:23:10 GMT",
            "x-amzn-requestid": "f7a74106-0bfd-44c4-83a9-0b8b27a8673f"
        },
        "http_status_code": 400,
        "request_id": "f7a74106-0bfd-44c4-83a9-0b8b27a8673f",
        "retry_attempts": 0
    }

```
To make this failures more easier to understand to the user I added the conditional check above which will throw this error in the same situation.
```
"msg": "Detaching instance(s) with 'decrement_desired_capacity' flag set reduces number of instances 
below min_size, please update AutoScalingGroup Sizes properly."
```

plugins/modules/ec2_asg.py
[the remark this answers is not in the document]
alinabuzachis: I was suggesting to ignore the instances inside the detach_instances that does not belong to any group (instances), and update the list you pass to the call with the only ones really belonging. In this way, you won't have this error.

plugins/modules/ec2_asg.py
alinabuzachis: Yes, I understood that. The thing confusing for me was the check. I would probably do something like this instead len(instances) - len(detach_instances) < min_size. But I mean, as you prefer.

alinabuzachis (commented):
@mandar242 You also would need a changelog fragment and integration tests for these options.

plugins/modules/ec2_asg.py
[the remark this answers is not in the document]
jillr: Agreed, let's check what instances are currently attached to the ASG and only detach the ones that exist.  We don't need to surface any errors to the user if a specific instance is already detached.

plugins/modules/ec2_asg.py
```diff
@@ -1523,6 +1544,40 @@ def replace(connection):
...
+        if min_size and min_size > decremented_desired_capacity:
+            module.fail_json(
+                msg="Detaching instance(s) with 'decrement_desired_capacity' flag set reduces number of instances below min_size,\
+                        please update AutoScalingGroup Sizes properly.")
+
+    if instances_to_detach:
+        try:
+            detach_asg_instances(connection, instances_to_detach, group_name, decrement_desired_capacity)
+        except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
+            module.fail_json_aws(e, msg="Failed to describe launch configurations")
```
jillr, line 1575:
```suggestion
            module.fail_json_aws(e, msg="Failed to detach instances from AutoScaling Group")
```

plugins/modules/ec2_asg.py
```diff
@@ -1523,6 +1544,40 @@ def replace(connection):
...
+            instances_to_detach.append(instance_id)
+
+    # check if setting decrement_desired_capacity will make desired_capacity smaller
+    # than the currently set minimum size in ASG configuration
+    if decrement_desired_capacity:
+        decremented_desired_capacity = len(instances) - len(instances_to_detach)
+        if min_size and min_size > decremented_desired_capacity:
+            module.fail_json(
+                msg="Detaching instance(s) with 'decrement_desired_capacity' flag set reduces number of instances below min_size,\
+                        please update AutoScalingGroup Sizes properly.")
```
jillr, line 1569: It would be nice to include here what the min_size is and how much we would have reduced desired capacity by, so the user knows how much they would need to reduce the size by.

jillr (commented):
Two minor nits on some error messages, otherwise the code looks good - thanks @mandar242

jillr:
recheck

jillr:
afaict the timeout is just because we have to wait for instances on several tasks and the API was being slow when the tests ran.

mandar242:
recheck

plugins/modules/ec2_asg.py
```diff
@@ -182,6 +182,21 @@
         matching the current launch configuration.
     type: list
     elements: str
+  detach_instances:
+    description:
+      - Removes one or more instances from the specified AutoScalingGroup.
+      - If I(decrement_desired_capacity) flag is not set, new instance(s) are lauched to replace the detached instance(s).
+      - If a Classic Load Balancer attached to the AutoScalingGroup, the instances are also deregistered from the load balancer.
```
jatorcasso, line 189:
```suggestion
      - If I(decrement_desired_capacity) flag is not set, new instance(s) are launched to replace the detached instance(s).
      - If a Classic Load Balancer is attached to the AutoScalingGroup, the instances are also deregistered from the load balancer.
```

plugins/modules/ec2_asg.py
```diff
@@ -1523,6 +1544,40 @@ def replace(connection):
     return changed, asg_properties
 
 
+def detach(connection):
+    group_name = module.params.get('name')
+    detach_instances = module.params.get('detach_instances')
+    as_group = describe_autoscaling_groups(connection, group_name)[0]
```
jatorcasso, line 1550: is there a case where multiple autoscaling groups are returned here?

jatorcasso (approved):
Left some minor grammar fixes - LGTM!

plugins/modules/ec2_asg.py
mandar242: I think as we are passing only single group name, we will only single auto scaling group's info.

jatorcasso: just checking! Wasnt sure if there were cases where multiple groups had the same name

jillr: ASG name is unique within a region and account.  Since we can only act on one region and account at a time, this works well.

jillr (approved):
The code lgtm, thanks very much @mandar242! I'll hold off on tagging this to merge over the weekend in case @alinabuzachis would like to update her review but with CI passing I think this is ready.

alinabuzachis (approved):

PR #943, opened 2022-02-21
rename cloudwatchevent_rule module to eventbridge_rule

themarcelor:
##### SUMMARY
I think the module handles EventBridge Rules and it is not related to CloudWatch 🤷🏼 

Ah, looks like there is some history [here](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-what-is.html):
> Note
> EventBridge was formerly called Amazon CloudWatch Events. The default event bus and the rules you created in CloudWatch Events also display in the EventBridge console. EventBridge uses the same CloudWatch Events API, so your code that uses the CloudWatch Events API stays the same. New features added to EventBridge are not added to CloudWatch Events.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
 eventbridge_rule

##### ADDITIONAL INFORMATION
It is a simple change:
```
mv plugins/modules/cloudwatchevent_rule.py plugins/modules/eventbridge_rule.py
sed -i 's/cloudwatchevent_rule/eventbridge_rule/' plugins/modules/eventbridge_rule.py
sed -i 's/CloudWatch/EventBridge/' plugins/modules/eventbridge_rule.py
```

That's probably a breaking change that will impact lots of people so... feel free to decline the PR.

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -52,7 +52,6 @@ action_groups:
   - cloudfront_invalidation
   - cloudfront_origin_access_identity
   - cloudtrail
-  - cloudwatchevent_rule
   - cloudwatchlogs_log_group
   - cloudwatchlogs_log_group_info
   - cloudwatchlogs_log_group_metric_filter
@@ -123,6 +122,7 @@ action_groups:
   - elb_target_group
   - elb_target_group_info
   - elb_target_info
+  - eventbridge_rule
   - execute_lambda
   - iam_access_key
   - iam_access_key_info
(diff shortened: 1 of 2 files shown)


markuman:
> That's probably a breaking change that will impact lots of people so... feel free to decline the PR.

yes :) You've answered everything by yourself.

Furthermore, it targets only one single module. But there are many more modules with nameing issues.  
A compete solution for the entire collection is desirable.

There is already an issue for that. Feel free to join and bring your point of view into it: https://github.com/ansible-collections/amazon.aws/issues/610

PR #482, opened 2021-03-17
Add ec2_asg_tag module

jsok:
##### SUMMARY

A new module to manage ASG tags, analogous to the `ec2_tag` module in
`amazon.aws` collection.

Fixes #481

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

`ec2_asg_tag`

##### ADDITIONAL INFORMATION

Motivation is outlined in #481, but this module provides a more declarative approach to managing ASG tags, e.g.:

```yaml
- name: Ensure tags are present on an ASG
  community.aws.ec2_asg_tag:
    resource: my-auto-scaling-group
    state: present
    tags:
      - environment: production
        propagate_at_launch: true
      - role: webserver
        propagate_at_launch: true

- name: Ensure tag is absent on an ASG
  community.aws.ec2_asg_tag:
    resource: my-auto-scaling-group
    state: absent
    tags:
      - environment: development
```

The module utilises underlying API calls to:
* https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_DescribeTags.html
* https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_CreateOrUpdateTags.html
* https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_DeleteTags.html

PR #960, opened 2022-03-02
ec2_asg: Add purge_tags to AutoScalingGroups.

### LINKED ISSUE #481: AutoScaling Group tags module
##### SUMMARY

Add a new module `ec2_asg_tag` analogous to `ec2_tag` to manage tags on AutoScaling Groups.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`ec2_asg_tag`

##### ADDITIONAL INFORMATION

Using `ec2_asg` to manage tags is difficult since you must [specify the entire list of tags to manage them](https://github.com/ansible-collections/community.aws/blob/9b60a2d5d50ff1d77e74cf79715ccebc4632757f/tests/integration/targets/ec2_asg/tasks/main.yml#L194-L248).
This is cumbersome since you first need to use `ec2_asg_info` to query the current set of tags, munge the result to create a new list of tags that includes or excludes the tags you with to add/remove and call `ec2_asg` again.

A more declarative approach would be to follow the pattern of `ec2_tag`, e.g.:

```yaml
- name: Ensure tags are present on an ASG
  community.aws.ec2_asg_tag:
    resource: my-auto-scaling-group
    state: present
    tags:
      - environment: production
        propagate_at_launch: true
      - role: webserver
        propagate_at_launch: true

- name: Ensure tag is absent on an ASG
  community.aws.ec2_asg_tag:
    resource: my-auto-scaling-group
    state: absent
    tags:
      - environment: development
```

The module would utilise underlying API calls to:
* https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_DescribeTags.html
* https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_CreateOrUpdateTags.html
* https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_DeleteTags.html

mandar242:
##### SUMMARY

Add ``purge_tags`` to ec2_asg module.

Fixes #481.
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ec2_asg

##### ADDITIONAL INFORMATION

There was another PR (currently closed) #482 - with similar functionality but I'm not sure if having modules to handle tags for individual services/modules is a good way to have this functionality. It will certainly cause increase in number of modules.
Hence tried modifying existing ```ec2_asg``` module to be able to do this.

This utilizes underlying API calls to:

https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_DescribeTags.html
https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_CreateOrUpdateTags.html
https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_DeleteTags.html


mandar242:
recheck

plugins/modules/ec2_asg.py
```diff
@@ -220,6 +220,25 @@
...
+      - If the I(tags) parameter is not set then tags will not be modified.
+    default: true
+    type: bool
+    version_added: 3.2.0
+  remove_tags:
+    description:
+      - If C(true), the tags specified by I(tags) parameter will be removed/deleted from AutoScalingGroup.
+      - If the I(tags) parameter is not set then tags will not be modified.
+    type: bool
+    version_added: 3.2.0
```
alinabuzachis, line 241: The question here is: do we want to keep a dedicated to module for doing this or handle tags directly here? @tremble @jillr @markuman If we fall into the handling of tags within this module, the only parameter you need is `purge_tags`. Conversely, if we agree with a new module, you should resurrect the original PR https://github.com/ansible-collections/community.aws/pull/482. 

```suggestion
  purge_tags:
    description:
      - If C(true), existing tags will be purged from the resource to match exactly what is defined by I(tags) parameter.
    default: true
    type: bool
    version_added: 3.2.0
```

alinabuzachis (changes requested):

plugins/modules/ec2_asg.py
jillr: I don't want a dedicated tags module.  We have ec2_tag and that's fine, but I don't want us to go down the path of a dedicated tag module for every service, either for our own maintenance or for users having to juggle so many modules.  I agree with @alinabuzachis about the diff here as well, CUD modules (like ec2_asg) in amazon.aws should generally not do list-only actions, ec2_asg_info returns tag info for this purpose.  We should just have the same purge_tags behaviour that we do in other, existing modules.
Thanks for your work on this @mandar242!

tests/integration/targets/ec2_asg/tasks/tag_operations.yml
```diff
@@ -0,0 +1,143 @@
...
+
+    - name: Get asg info
+      ec2_asg_info:
+        name: "{{ resource_prefix }}-asg-tag-test"
+      register: info_result
+
+    - assert:
+        that:
+          - add_result is changed
+          - info_result.results[0].tags | length == 4
```
alinabuzachis, line 77: Can you also check tag values overall?

tests/integration/targets/ec2_asg/tasks/tag_operations.yml
```diff
@@ -0,0 +1,143 @@
...
+      ec2_asg:
+        name: "{{ resource_prefix }}-asg-tag-test"
+        tags:
+          - key1: 'val1'
+            propagate_at_launch: no
+          - key2: 'val2'
+            propagate_at_launch: yes
+        purge_tags: false
+      register: add_result
+
```
alinabuzachis, line 111: Can you also add another task with `tags: []` with  `purge_tags: false` and check if the existing tags are retained?

tests/integration/targets/ec2_asg/tasks/tag_operations.yml
```diff
@@ -0,0 +1,143 @@
...
+    - name: Add 2 new tags - purge existing tags
+      ec2_asg:
+        name: "{{ resource_prefix }}-asg-tag-test"
+        tags:
+          - tag_a: 'val_a'
+            propagate_at_launch: no
+          - tag_b: 'val_b'
+            propagate_at_launch: yes
+      register: add_purge_result
+
```
alinabuzachis, line 89: Please add another task that update the tags (remove and add).

tests/integration/targets/ec2_asg/tasks/tag_operations.yml
```diff
@@ -0,0 +1,143 @@
...
+        name: "{{ resource_prefix }}-asg-tag-test"
+      register: info_result
+
+    - assert:
+        that:
+          - add_result is changed
+          - info_result.results[0].tags | length == 4
+          - '"autoscaling:CreateOrUpdateTags" in add_result.resource_actions'
+
+    #----------------------------------------------------------------------
```
alinabuzachis, line 123: Please also consider testing with similar tags:

```
tags:
        lowercase spaced: "hello cruel world"
        Title Case: "Hello Cruel World"
        CamelCase: "SimpleCamelCase"
        snake_case: "simple_snake_case"
```

tests/integration/targets/ec2_asg/tasks/tag_operations.yml
```diff
@@ -0,0 +1,250 @@
...
+    - assert:
+        that:
+          - add_empty is not changed
+          - info_result.results[0].tags | length == 4
+          - '"tag_a" in tag_keys'
+          - '"tag_b" in tag_keys'
+          - '"lowercase spaced" in tag_keys'
+          - '"Title Case" in tag_keys'
+          - '"autoscaling:CreateOrUpdateTags" not in add_empty.resource_actions'
+
```
alinabuzachis, line 229: I tried to completely remove the existing tags with, but it didn't remove anything. Can you fix that @mandar242 ?
```
    - name: Add empty tags with purge set to True
      ec2_asg:
        name: "{{ resource_prefix }}-asg-tag-test"
        tags: []
      register: add_empty
```

alinabuzachis (changes requested):

tests/integration/targets/ec2_asg/tasks/tag_operations.yml
mandar242: @alinabuzachis Thank you for pointing this out!
Fixed.

alinabuzachis (approved):

plugins/modules/ec2_asg.py
```diff
@@ -220,6 +220,13 @@
       - When I(propagate_at_launch) is true the tags will be propagated to the Instances created.
     type: list
     elements: dict
+  purge_tags:
+    description:
+      - If C(true), existing tags will be purged from the resource to match exactly what is defined by I(tags) parameter.
+      - If the I(tags) parameter is not set then tags will not be modified.
+    default: true
```
bedge, line 227: Forgive me for commenting on an already merged PR, but I just found out about this from a complete breakage of all ansible AWS services that use tags.

Why was `default=True` chosen, for a new field that no one was using previously, deemed a good idea?

Now, everything that touches any asset that uses tags MUST now add a `purge_tags: False` just to leave things as they are??

Otherwise all tags are deleted???

Existing use case to scale in/out an ASG, and touch nothing else:
```
- community.aws.ec2_asg:
    name: "{{ item.auto_scaling_group_name }}"
    min_size: "{{ count }}"
    max_size: "{{ count }}"
    desired_capacity: "{{ count }}"
    region: eu-central-1
```

Now, the above and strips all tags as well. 

Given that most ASG tags are used to propagate context information to the ec2 instances, this now destroys all context information for the ASG and the instances it creates have no information about the runtime context.

How is that the "principle of least surprising behavior"?

This now requires adding the purge_tags: False everywhere where tags could be impacted.
```
- community.aws.ec2_asg:
    name: "{{ item.auto_scaling_group_name }}"
    min_size: "{{ count }}"
    max_size: "{{ count }}"
    desired_capacity: "{{ count }}"
    purge_tags: False       <----- This is now needed everywhere tags are used to say "leave alone" ?
    region: eu-central-1
```

This is the ONLY case where one has to specify a non-default option just to say: "leave alone".

Note also that
```
    purge_tags: False 
```
is NOT backwards compatible, because that field didn't exist in previous versions, so you need 2 different cases depending on which side of this version you fall on.

Given that the `purge_tags` things seems to have propagated to many AWS assets, all in slightly different releases, one needs to account for this on a case by case basis depending on when the `purge_tags` was introduced into each module.

Now, if the default had been `purge_tags=False`, none of this would be an issue. 
There would have been no changes required as the new behavior would match the old.

Am I the only one complaining about this? I can't believe this didn't hit more people.
Does no use use ansible to alter "only what is specified" and expect the rest to remain as-is?

I can't decide now whether to walk away from ansible as an AWS "minor edit" tool or, try deal with this "maybe we'll change something that destroys something else later, that you can't know about until it happens.

jillr:
Hi @bedge.  I'm sorry to hear that this change caused unexpected behaviour in your environment, that's definitely not the experience we want users to have.  `purge_tags=True` is the default in a majority of Ansible AWS modules that support tagging and we did communicate the new option in the [changelog](https://github.com/ansible-collections/community.aws/blob/3.2.0/CHANGELOG.rst#minor-changes) for the collection update.  That doesn't mean that this setting is perfect or set in stone though - we would be happy to take a bug report in a new issue where we can discuss as a community if we chose the correct default here and if it should be changed.

The developers who work on this collection are human and while we try our best sometimes humans make mistakes, or we make decisions without the same context that someone else has.  I can assure you that no one is setting out to "change something that destroys something else later, that you can't know about until it happens".  We strive to communicate all changes in a release in the changelog, if there's something we could have done differently with that changelog entry we would also welcome that feedback in the bug report.

PR #986, opened 2022-03-14
Prevent CI to be run for whole collection when modules without integration tests are modified

alinabuzachis:
Add a fake integration suite including an aliases file listing every module name with missing integration tests. 

This fake suite is necessary for the new CI ansible-test-splitter behaviour. Namely, if one of the modules (listed in the aliases file) without a test suite is modified, the CI is run for the entire collection since the ansible-test-splitter won't find any target match. This fake integration suite helps handle this situation by avoiding running the CI for the whole collection. Furthermore, since the modules listed in the aliases file are marked as disabled, tests are automatically skipped).

Add some of the modules in the aliases files where it makes more sense.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


tremble (approved):
Generally LGTM, probably also worth adding a pinned Issue with a ToDo list for these (I think an issue per-missing test would be a little spammy).

alinabuzachis:
> Generally LGTM, probably also worth adding a pinned Issue with a ToDo list for these (I think an issue per-missing test would be a little spammy).

Thanks! Opened here https://github.com/ansible-collections/community.aws/issues/988

PR #1002, opened 2022-03-17
rds_instance - fix check_mode and idempotence bugs and support adding/removing iam roles

### LINKED ISSUE #464: rds_instance module should support adding IAM roles
##### SUMMARY
AWS supports adding IAM roles to RDS instance. It is required for an RDS instance to be able to access an S3 bucket.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
rds_instance

##### ADDITIONAL INFORMATION
AWS CLI supports this via
https://docs.aws.amazon.com/cli/latest/reference/rds/add-role-to-db-instance.html

Adding S3 integration to RDS is described in 
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/oracle-s3-integration.html#oracle-s3-integration.preparing

There is an article on Stack Overflow describing a workaround with plays using AWS CLI to do this
https://stackoverflow.com/questions/63010540/attaching-iam-role-to-aws-rds-instance-with-ansible

I think most people would set this up when creating a new RDS instance, so it would make sense
to add this functionality to the rds_instance module instead of making another module.

Since AWS CLI takes feature-name and role-arn arguments, perhaps the usage could be as follows:
```yaml
rds_instance:
  db_instance_identifier: 'test-instance'
  iam_roles:
    - role_arn: 'arn:aws:iam::111122223333:role/rds-s3-integration-role'
      feature_name: 'S3_INTEGRATION'
```

### LINKED ISSUE #1013: rds_instance - Broken check_mode and idempotence
### Summary

When adding integration tests for adding iam_roles support for rds_instance, I noticed check_mode tests fail on idempotence checks, as well as some other checks like adding sgs on check_mode, etc.

### Issue Type

Bug Report

### Component Name

rds_instance

### Ansible Version

```console (paste below)
$ ansible --version

```

### Collection Versions

```console (paste below)
$ ansible-galaxy collection list
```

### AWS SDK versions

```console (paste below)
$ pip show boto boto3 botocore
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### OS / Environment

_No response_

### Steps to Reproduce

```yaml (paste below)
    - name: Create DB instance with IAM roles (idempotence) - check_mode
      rds_instance:
        id: "{{ instance_id }}"
        state: present
        engine: postgres
        engine_version: "{{ postgres_db_engine_version }}"
        username: "{{ username }}"
        password: "{{ password }}"
        db_instance_class: "{{ postgres_db_instance_class }}"
        allocated_storage: "{{ allocated_storage }}"
        allow_major_version_upgrade: yes
        iam_roles:
          - role_arn: "{{ s3_integration_role_1.arn }}"
            feature_name: 's3Export'
          - role_arn: "{{ s3_integration_role_2.arn }}"
            feature_name: 'Lambda'
          - role_arn: "{{ s3_integration_role_3.arn }}"
            feature_name: 's3Import'
      register: result
      check_mode: yes

    - assert:
        that:
          - not result.changed
```

### Expected Results

expected check_mode to not modify anything and return the correct `changed`

### Actual Results

in idempotence tests, check_mode returned `changed = True`. When adding sg on check_mode, the sgs were actually added.

### Code of Conduct

jatorcasso:
##### SUMMARY
- Support the addition and deletion of iam roles to db instances
- Fixes #464 
- Fixes #1013 
- Integration tests to test both this and the amazon.aws module_util rds changes

Depends-On https://github.com/ansible-collections/amazon.aws/pull/714

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
rds_instance

##### ADDITIONAL INFORMATION
Wasn't sure the best way to go about deleting IAM roles - ended up using a `purge_iam_roles` param that defaults to `False`, which seems consistent with other modules I've looked at.


tests/integration/targets/rds_instance/inventory
```diff
@@ -14,6 +14,7 @@ modify
 tagging
 replica
 upgrade
+iam_roles
```
marknet15, line 18: You may also want to increase this to allow the extra tests to run in parallel:
https://github.com/ansible-collections/community.aws/blob/main/tests/integration/targets/rds_instance/main.yml#L9

The RDS tests are very very slow !

markuman:
Not all RDS engines supports IAM roles. E.g. mariadb.  
Do we plan to cover this?

jatorcasso:
> Not all RDS engines supports IAM roles. E.g. mariadb. Do we plan to cover this?

That's a good point. I suppose that can be added as a pre-check in rds_instance or as a caught error in amazon.aws.module_utils.rds.handle_errors. I would guess add it to the utils handle_errors since that already exists, but im never really sure which collection I should be modifying..

alinabuzachis:
> > Not all RDS engines supports IAM roles. E.g. mariadb. Do we plan to cover this?
> 
> That's a good point. I suppose that can be added as a pre-check in rds_instance or as a caught error in amazon.aws.module_utils.rds.handle_errors. I would guess add it to the utils handle_errors since that already exists, but im never really sure which collection I should be modifying..

I guess a pre-check inside rds_instance will be enough.

jatorcasso:
> > > Not all RDS engines supports IAM roles. E.g. mariadb. Do we plan to cover this?
> > 
> > 
> > That's a good point. I suppose that can be added as a pre-check in rds_instance or as a caught error in amazon.aws.module_utils.rds.handle_errors. I would guess add it to the utils handle_errors since that already exists, but im never really sure which collection I should be modifying..
> 
> I guess a pre-check inside rds_instance will be enough. Please also document the engines supporting aim roles.

I actually think it makes more sense in this case to add an extra conditional in the `handle_errors` method, similar to what is done for when the user passes in an invalid engine. something like:

```
    elif method_name == 'add_role_to_db_instance' and error_code == 'InvalidParameterValue':
        if 'is not valid for the engine' in to_text(exception) and parameters.get('Engine') not in accepted_engines:
            module.fail_json_aws(exception, msg='It appears you are trying to modify IAM roles for an engine that does not support IAM roles.')
        else:
            module.fail_json_aws(exception, msg='Unable to {0}'.format(get_rds_method_attribute(method_name, module).operation_description))
```

Only issue is the accepted engines here are not well documented

alinabuzachis:
> > > > Not all RDS engines supports IAM roles. E.g. mariadb. Do we plan to cover this?
> > > 
> > > 
> > > That's a good point. I suppose that can be added as a pre-check in rds_instance or as a caught error in amazon.aws.module_utils.rds.handle_errors. I would guess add it to the utils handle_errors since that already exists, but im never really sure which collection I should be modifying..
> > 
> > 
> > I guess a pre-check inside rds_instance will be enough. Please also document the engines supporting aim roles.
> 
> I actually think it makes more sense in this case to add an extra conditional in the `handle_errors` method, similar to what is done for when the user passes in an invalid engine. something like:
> 
> ```
>     elif method_name == 'add_role_to_db_instance' and error_code == 'InvalidParameterValue':
>         if 'is not valid for the engine' in to_text(exception) and parameters.get('Engine') not in accepted_engines:
>             module.fail_json_aws(exception, msg='It appears you are trying to modify IAM roles for an engine that does not support IAM roles.')
>         else:
>             module.fail_json_aws(exception, msg='Unable to {0}'.format(get_rds_method_attribute(method_name, module).operation_description))
> ```
> 
> Only issue is the accepted engines here are not well documented

Where did you find that ham roles are not supported by all the engines https://boto3.amazonaws.com/v1/documentation/api/1.9.185/reference/services/rds.html#RDS.Client.add_role_to_db_instance? This only happens for rds_cluster.

alinabuzachis:
> its convoluted, but some engines dont support any features. for example from https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DBEngineVersion.html:
> 
> > SupportedFeatureNames.member.N
> > A list of features supported by the DB engine.
> > The supported features vary by DB engine and DB engine version.
> > To determine the supported features for a specific DB engine and DB engine version using the AWS CLI, use the following command:
> > aws rds describe-db-engine-versions --engine <engine_name> --engine-version <engine_version>
> > For example, to determine the supported features for RDS for PostgreSQL version 13.3 using the AWS CLI, use the following command:
> > aws rds describe-db-engine-versions --engine postgres --engine-version 13.3
> > The supported features are listed under SupportedFeatureNames in the output.
> > Type: Array of strings
> > Required: No
> 
> I had to learn this in order to properly test.. many of the db engines simply return an empty list of supported features.

Ok. yeah! So I guess you need to handle that as you suggested directly inside module_utils in amazon.aws. Cover it by a unit test.

markuman:
> Where did you find that iam roles are not supported by all the engines ... 

... just experiences :)

markuman:
> Ok. yeah! So I guess you need to handle that as you suggested directly inside module_utils in amazon.aws. Cover it by a unit test.

I'm not sure .... if `rds_instance` is moved soon into amazon.aws, ok. otherwise feels easier to maintain those technical deprecation in `rds_instance` itself.  
Because when AWS maybe support this one e.g. on 1st april, you need to wait for a fix in amazon.aws and just backport might be also not that easy.

jatorcasso:
https://awscli.amazonaws.com/v2/documentation/api/latest/reference/rds/describe-db-engine-versions.html

looks like the valid db engines need to be updated as well. I can make that a separate issue/corresponding PR?

jatorcasso:
> > Ok. yeah! So I guess you need to handle that as you suggested directly inside module_utils in amazon.aws. Cover it by a unit test.
> 
> I'm not sure .... if `rds_instance` is moved soon into amazon.aws, ok. otherwise it feels easier to maintain those technical deprecation in `rds_instance` itself. Because when AWS maybe support this one e.g. on 1st april, you need to wait for a fix in amazon.aws and just backport might be also not that easy.

I'll add in `rds_instance`. It's not much anyway

markuman:
> https://awscli.amazonaws.com/v2/documentation/api/latest/reference/rds/describe-db-engine-versions.html
> 
> looks like the valid db engines need to be updated as well. I can make that a separate issue/corresponding PR?

If we want to land this in 3.2.0, it might be easier to include the updated engine list also in this PR. But it's upt to you.

jatorcasso:
Marked as draft to include changes for https://github.com/ansible-collections/community.aws/issues/1013

plugins/modules/rds_instance.py
```diff
@@ -942,23 +1017,21 @@ def get_current_attributes_with_inconsistent_keys(instance):
...
     return options
 
 
 def get_changing_options_with_inconsistent_keys(modify_params, instance, purge_cloudwatch_logs, purge_security_groups):
     changing_params = {}
     current_options = get_current_attributes_with_inconsistent_keys(instance)
-
-    if current_options.get("MaxAllocatedStorage") is None:
-        current_options["MaxAllocatedStorage"] = None
-
```
jatorcasso, line 961: removed this because `MaxAllocatedStorage` is checked in `get_changing_options_with_consistent_keys`. fixes idempotence bug in test_complex

jatorcasso:
> > https://awscli.amazonaws.com/v2/documentation/api/latest/reference/rds/describe-db-engine-versions.html
> > looks like the valid db engines need to be updated as well. I can make that a separate issue/corresponding PR?
> 
> If we want to land this in 3.2.0, it might be easier to include the updated engine list also in this PR. But it's upt to you.

I updated here by adding `choices` list to engine parameter

markuman:
recheck

jatorcasso:
marking as draft for the time being - blocked by lack of permissions but how to handle is an open discussion atm

plugins/modules/rds_instance.py
```diff
@@ -982,8 +1052,13 @@ def get_changing_options_with_inconsistent_keys(modify_params, instance, purge_c
                 if desired_option in current_option:
                     continue
 
-        if current_option == desired_option:
-            continue
+        if option != 'ProcessorFeatures':
```
goneri, line 1055: Nitpicking and please don't refactor your code just because of my feeling. I think these two blocks are a bit hard to read, I would flatten the if/else this way.

```python
if option != 'ProcessorFeatures' and current_option == desired_option:
    continue

if option == 'ProcessorFeatures' and current_option == boto3_tag_list_to_ansible_dict(desired_option, 'Name', 'Value'):
    continue
```

goneri (approved):

tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_upgrade.yml
```diff
@@ -62,6 +62,46 @@
...
         allocated_storage: "{{ allocated_storage }}"
+        apply_immediately: True
+      register: result
+
+    - assert:
+        that:
+          - result.changed
+          - '"engine_version" in result.pending_modified_values or result.engine_version == mariadb_engine_version_2'
+
+    ### Specifying allow_major_version_upgrade with check_mode will always result in changed=True
```
alinabuzachis, line 73: Your aren't specifying `allow_major_version_upgrade`.

plugins/modules/rds_instance.py
```diff
@@ -1239,17 +1351,30 @@ def main():
...
 
     if method_name:
+
+        # Exit on create/delete if check_mode
+        if module.check_mode and method_name in ['create_db_instance', 'delete_db_instance']:
+            module.exit_json(changed=True, **camel_dict_to_snake_dict(instance, ignore_list=['Tags', 'ProcessorFeatures']))
+
         raw_parameters = arg_spec_to_rds_params(dict((k, module.params[k]) for k in module.params if k in parameter_options))
-        parameters = get_parameters(client, module, raw_parameters, method_name)
+        parameters_to_modify = get_parameters(client, module, raw_parameters, method_name)
```
alinabuzachis, line 1361: I would probably also add a warning when the user sets `allow_major_version_upgrade` parameter with a descriptive message ('The idempotency is not guaranteed whith allow_major_version_upgrade' or something that makes more sense).

tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_upgrade.yml
jatorcasso: That was intentional. The comment there was saying "hey, I'm not testing allow_major_version_upgrade, because if I did, check mode idempotence test would fail"

plugins/modules/rds_instance.py
jatorcasso: So I was going to do this, but it only fails on check mode (since the rds util will catch and handle no changes _actually_ being made and return changed=False). I guess I can add the warning when it's specified during a check mode run

tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_upgrade.yml
alinabuzachis: But I guess it more clear if you remove the assert and use  `allow_major_version_upgrade ` as a parameter and leave the comment explaining the issue or you can comment the expected assert - not changed.

plugins/modules/rds_instance.py
alinabuzachis: Not sure I get the problem. I guess something like this should work:

if modules.params.get(allow_major_version_upgrade):
    module.warns('Exhaustive Message here')

jatorcasso: the problem only exists when run in check mode - ill update with a warning in that scenario. thanks!

tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_upgrade.yml
jatorcasso: that makes sense. ill update thanks

alinabuzachis (approved):

jatorcasso:
@markuman @jillr I supposed I should change version_added to 3.3.0 and add backport-3 label ?

markuman (approved):

PR #1026, opened 2022-03-28
Add AWSRetry.jittered_backoff to rds_instance_info

kteague-tasktop:
##### SUMMARY

Add AWSRetry.jittered_backoff to the rds_instance_info module.

When calling `rds_instance_info` we have been seeing API rate limit errors from AWS. When calling this module, it usually runs to about 90-150 times in a minute before we get rate limited.

Using jittered_backoff should significantly decrease the number of times we see API rate limits here.

```
02:20:36 An exception occurred during task execution. To see the full traceback, use -vvv. The error was: botocore.exceptions.ClientError: An error occurred (Throttling) when calling the DescribeDBInstances operation (reached max retries: 4): Rate exceeded
02:20:36 fatal: [polaris -> localhost]: FAILED! => {"boto3_version": "1.20.22", "botocore_version": "1.23.22", "changed": false, "error": {"code": "Throttling", "message": "Rate exceeded", "type": "Sender"}, "msg": "Couldn't get instance information: An error occurred (Throttling) when calling the DescribeDBInstances operation (reached max retries: 4): Rate exceeded", "response_metadata": {"http_headers": {"connection": "close", "content-length": "254", "content-type": "text/xml", "date": "Tue, 15 Mar 2022 09:20:34 GMT", "x-amzn-requestid": "5de8131e-3f59-4b04-af25-5f7083ee09b9"}, "http_status_code": 400, "max_attempts_reached": true, "request_id": "5de8131e-3f59-4b04-af25-5f7083ee09b9", "retry_attempts": 4}}
```

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
- rds_instance_info

##### ADDITIONAL INFORMATION
Decorated rds_instance_info with AWSRetry.jittered_backoff

### DIFF
--- a/plugins/modules/rds_instance_info.py
+++ b/plugins/modules/rds_instance_info.py
@@ -365,6 +365,17 @@ except ImportError:
     pass  # handled by AnsibleAWSModule
 
 
+@AWSRetry.jittered_backoff()
+def _describe_db_instances(conn, **params):
+    paginator = conn.get_paginator('describe_db_instances')
+    try:
+        results = paginator.paginate(**params).build_full_result()['DBInstances']
+    except is_boto3_error_code('DBInstanceNotFound'):
+        results = []
+
+    return results
+
+
 def instance_info(module, conn):
     instance_name = module.params.get('db_instance_identifier')
     filters = module.params.get('filters')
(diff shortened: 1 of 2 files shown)


markuman:
recheck

plugins/modules/rds_instance_info.py
```diff
@@ -359,6 +359,7 @@
     pass  # handled by AnsibleAWSModule
 
 
+@AWSRetry.jittered_backoff()
```
marknet15, line 362: I can't quite remember if its enough for use with the paginator to just add this, what I normally do is split out the paginator like this into its own function so that wrapped up calls still get covered:
https://github.com/ansible-collections/community.aws/blob/main/plugins/modules/elb_target_group_info.py#L220

marknet15:
Hey @kteague-tasktop at the moment as there are no set tests for the module plugin, the whole test suite is running covering all modules. If it's not too much trouble you could create a test target here for `rds_instance_info`:
https://github.com/ansible-collections/community.aws/tree/main/tests/integration/targets
it would just need a very small subset of the tasks from here:
https://github.com/ansible-collections/community.aws/tree/main/tests/integration/targets/rds_instance

to create a RDS instance then a call to `rds_instance_info` to verify it's functionality.  normally the info and non info tests are bundled but in this case RDS related ops take a long time to run !

tremble (changes requested):
Thanks for this.  At the least we need some minimal integration tests.

Would you be able to update tests/integration/targets/rds_instance/ to include some basic tests of the _info module?  they're just playbooks/roles that run in the CI account to test the functionality of modules

markuman:
@tremble @marknet15 can we merge this now?

PR #1035, opened 2022-03-30
prepare 2.4.0

markuman:
##### SUMMARY

Release of 2.4.0

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME

entire collection

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -8,7 +8,7 @@ description: null
 license_file: COPYING
 tags: [community, aws, cloud, amazon]
 dependencies:
-  amazon.aws: '>=2.2.0'
+  amazon.aws: '>=2.2.0,<3.0.0'
 repository: https://github.com/ansible-collections/community.aws
 documentation: https://github.com/ansible-collections/community.aws/tree/main/docs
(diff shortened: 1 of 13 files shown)


galaxy.yml
```diff
@@ -1,6 +1,6 @@
 namespace: community
 name: aws
-version: 2.3.0
+version: 2.4.0
 readme: README.md
 authors:
   - Ansible (https://github.com/ansible)
```
alinabuzachis, line 6: Do we need to update dependencies with amazon.aws 2.2.0? Or amazon.aws >= 2.1.0 is fine?

alinabuzachis (approved):

galaxy.yml
markuman: I don't know of any new dependencies that were introduced. In theory, since 3.x.x is released, the 2.x.x releases should receive only bugfixes and docfixed.  
But I'm fine to increase also the minimum `amazon.aws` requirements, because it introduces only bugixes :)

marknet15 (approved):

PR #1048, opened 2022-04-05
ec2_asg_lifecycle_hook: add integration tests 

mandar242:
##### SUMMARY

Adding integration tests to ec2_asg_lifecycle_hook module.
Fixing idempotency.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ec2_asg_lifecycle_hook

### DIFF
--- a/plugins/modules/ec2_asg_lifecycle_hook.py
+++ b/plugins/modules/ec2_asg_lifecycle_hook.py
@@ -100,16 +100,17 @@ RETURN = '''
 
 '''
 
-from ansible_collections.amazon.aws.plugins.module_utils.core import AnsibleAWSModule
 
 try:
     import botocore
 except ImportError:
     pass  # handled by AnsibleAWSModule
 
+from ansible_collections.amazon.aws.plugins.module_utils.core import AnsibleAWSModule
+from ansible.module_utils.common.dict_transformations import camel_dict_to_snake_dict
+
 
 def create_lifecycle_hook(connection, module):
-    changed = False
 
     lch_name = module.params.get('lifecycle_hook_name')
     asg_name = module.params.get('autoscaling_group_name')
@@ -119,6 +120,9 @@ def create_lifecycle_hook(connection, module):
     notification_meta_data = module.params.get('notification_meta_data')
     heartbeat_timeout = module.params.get('heartbeat_timeout')
     default_result = module.params.get('default_result')
+
+    return_object = {}
+    return_object['changed'] = False
 
     lch_params = {
         'LifecycleHookName': lch_name,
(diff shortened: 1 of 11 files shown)


jillr:
recheck

plugins/modules/ec2_asg_lifecycle_hook.py
```diff
@@ -150,23 +154,30 @@ def create_lifecycle_hook(connection, module):
...
             module.fail_json_aws(e, msg="Failed to create LifecycleHook")
 
-    return(changed)
+    else:
+        added, removed, modified, same = dict_compare(lch_params, existing_hook[0])
+        if modified:
+            # GlobalTimeout is not configurable, but exists in response.
+            # Removing it helps to compare both dicts in order to understand
+            # what changes were done.
+            del(existing_hook[0]['GlobalTimeout'])
```
jatorcasso, line 171: I get the logic here for only reporting changed on `modified`, but in that case it looks like you can get rid of the `del` (and the comments with it), since it's not really being used

mandar242: @jatorcasso yes,that makes sense, as I'm just trying to add integration tests here, didn't wanted to make any major changes to the module unless required. 
I can go ahead and remove it if @alinabuzachis or @jillr thinks that's a good idea.

plugins/modules/ec2_asg_lifecycle_hook.py
```diff
@@ -150,23 +154,30 @@ def create_lifecycle_hook(connection, module):
...
+            try:
+                return_object['changed'] = True
+                connection.put_lifecycle_hook(**lch_params)
+                return_object['lifecycle_hook_info'] = connection.describe_lifecycle_hooks(
+                    AutoScalingGroupName=asg_name, LifecycleHookNames=[lch_name])['LifecycleHooks']
+            except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
+                module.fail_json_aws(e, msg="Failed to create LifecycleHook")
+
+    module.exit_json(**camel_dict_to_snake_dict(return_object))
```
jillr, line 180: I like that we're returning the object now, but we should to document the values that are returned in the RETURN block of the documentation and include a comment in the changelog that the module now returns information about the LCH, so users can update their playbooks if they want to take advantage of this data.

jillr (approved):
Good work @mandar242, just a small docs request then this LGTM

plugins/modules/ec2_asg_lifecycle_hook.py
jillr: If the code is not needed we should remove it so that it has less chance of introducing a big later.  This small of a code change is fine I think.

jatorcasso (approved):
LGTM!

PR #1061, opened 2022-04-13
ec2_asg: minor fix - run setup only once

mandar242:
##### SUMMARY

Fix integration tests bug in ec2_asg and run env_setup only once.

The bug causes env_setup.yml to run as many times as the number of tests in inventory file.
This causes removing test VPC to fail with the error
`
"msg": "Currently there are 3 VPCs that have the same name and CIDR block you specified. If you would like to create the VPC anyway please pass True to the multi_ok param.",
`

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ec2_asg

### DIFF
--- a/tests/integration/targets/ec2_asg/roles/ec2_asg/tasks/env_setup.yml
+++ b/tests/integration/targets/ec2_asg/roles/ec2_asg/tasks/env_setup.yml
@@ -1,5 +1,5 @@
 - name: Run ec2_asg integration tests.
-
+  run_once: '{{ ec2_asg_setup_run_once }}'
   block:
 
     # ============================================================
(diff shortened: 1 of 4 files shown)


jatorcasso (approved):
LGTM after changelog

tests/integration/targets/ec2_asg/roles/ec2_asg/tasks/env_setup.yml
```diff
@@ -1,5 +1,5 @@
 - name: Run ec2_asg integration tests.
-
+  run_once: '{{ setup_run_once | default("no") | bool }}'
```
goneri, line 2: I would instead use `tests/integration/targets/ec2_asg/roles/ec2_asg/default/main.yaml` to define a default `setup_run_once` value for the role.

goneri (approved):

tests/integration/targets/ec2_asg/roles/ec2_asg/tasks/env_setup.yml
mandar242: Done. Thanks!

PR #1064, opened 2022-04-13
ec2_asg: Change purge_tags default value to False

mandar242:
##### SUMMARY

Changed default value of `purge_tags` to `False`.

With the addition of `purge_tags` to `ec2_asg` module #960, the default value was kept to `True` similar to many other modules in this collection and also as specified in [ansible dev guide](https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/dev_guide/platforms/aws_guidelines.rst#dealing-with-tags).

With this addition, the issue was discovered that there is a possibility of this change breaking existing playbooks for users if they don't update their playbooks and specify `purge_tags: false` where required.

This PR's change will prevent accidental breakage.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ec2_asg

### DIFF
--- a/plugins/modules/ec2_asg.py
+++ b/plugins/modules/ec2_asg.py
@@ -224,7 +224,7 @@ options:
     description:
       - If C(true), existing tags will be purged from the resource to match exactly what is defined by I(tags) parameter.
       - If the I(tags) parameter is not set then tags will not be modified.
-    default: true
+    default: false
     type: bool
     version_added: 3.2.0
   health_check_period:
@@ -1861,7 +1861,7 @@ def main():
         wait_timeout=dict(type='int', default=300),
         state=dict(default='present', choices=['present', 'absent']),
         tags=dict(type='list', default=[], elements='dict'),
-        purge_tags=dict(type='bool', default=True),
+        purge_tags=dict(type='bool', default=False),
         health_check_period=dict(type='int', default=300),
         health_check_type=dict(default='EC2', choices=['EC2', 'ELB']),
         default_cooldown=dict(type='int', default=300),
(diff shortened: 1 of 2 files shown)


mandar242:
recheck

mandar242:
recheck

markuman:
@mandar242 @alinabuzachis @jillr  once this is merged, do we ship it with 3.2.1.  
otherwise it is a breaking change - even if it is not a worse one - imo.

tremble (approved):
Let's undo this breaking change, but also look at how we want to do this.  About 3 quarters of our modules that support managing tags default to purge_tags=True behaviour today (some even nuke tags if you don't set the tags parameter)

markuman:
> Let's undo this breaking change, but also look at how we want to do this. About 3 quarters of our modules that support managing tags default to purge_tags=True behaviour today (some even nuke tags if you don't set the tags parameter)

I put this to the agenda for the next meeting.  
So we're going to release 3.2.1 once it is merged?

markuman (approved):

marknet15 (approved):

markuman:
regate

felixfontein:
Shouldn't this have a changelog fragment so that users know what happens?

markuman:
> Shouldn't this have a changelog fragment so that users know what happens?

yeah. good point. It can be added in the release PR, which also needs to be approved and where some extra notes should be left in the `release_summary`

markuman:
I've pushed a changelog fragment. It needs some reviews again  @tremble  @marknet15  @felixfontein

tremble (approved):

alinabuzachis (approved):

PR #1076, opened 2022-04-19
aws_api_gateway_domain: wrap boto3 calls in try/except

markuman:
##### SUMMARY
wrap boto3 calls in try/except.

no changelog fragment because not released yet.  
completed from: https://github.com/ansible-collections/community.aws/pull/44#discussion_r816798638

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

aws_api_gateway_domain

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/aws_api_gateway_domain.py
+++ b/plugins/modules/aws_api_gateway_domain.py
@@ -125,8 +125,8 @@ def get_domain(module, client):
     domain_name = module.params.get('domain_name')
     result = {}
     try:
-        result['domain'] = get_domain_name(client, domain_name)
-        result['path_mappings'] = get_domain_mappings(client, domain_name)
+        result['domain'] = get_domain_name(client, domain_name, module)
+        result['path_mappings'] = get_domain_mappings(client, domain_name, module)
     except is_boto3_error_code('NotFoundException'):
         return None
     except (ClientError, BotoCoreError, EndpointConnectionError) as e:  # pylint: disable=duplicate-except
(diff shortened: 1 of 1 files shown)


stefanhorning:
@markuman thanks for finishing this up! :+1:

alinabuzachis (approved):

jatorcasso (approved):

plugins/modules/aws_api_gateway_domain.py
```diff
@@ -231,43 +231,58 @@ def delete_domain(module, client):
 
 
 @AWSRetry.backoff(**retry_params)
-def get_domain_name(client, domain_name):
-    return client.get_domain_name(domainName=domain_name)
+def get_domain_name(client, domain_name, module):
+    try:
+        return client.get_domain_name(domainName=domain_name)
+    except (BotoCoreError, ClientError) as e:
+        module.fail_json_aws(e, msg="Failed to get api gateway domain.")
```
s-hertel, line 238: It looks like this will fail the first time a ClientError is encountered rather than retrying. Generally the exception handling should happen outside the retry wrapper.

markuman: That's a good point.   
In that case it is already covered by try/catch L127-134.  
So I'm going to revert it and review the other cases also.

markuman: Ok, all client calls all already covered by try and catch :)  
So I am going to close this PR.

s-hertel: Thanks for checking :smile:

PR #1475, opened 2022-09-15
tests/ec2_vpc_egress_igw: avoid a InvalidVpcId.Malformed error

goneri:
Ensures the "test failure with non-existent VPC ID" test uses a properly
formatted VPC ID.
Otherwise, the API now returns the following error:

```
"error": {
    "code": "InvalidVpcId.Malformed",
    "message": "The vpc ID vpc-012345678 is malformed"
},
```


goneri:
regate

tremble (approved):

Issue #1266, opened 2022-06-25
EFS: Add backup option


nick-zh:
### Summary

Add the `Backup` option for EFS

### Issue Type

Feature Idea

### Component Name

efs

### Additional Information

From what i can see [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/efs.html#EFS.Client.create_file_system), i assume it should be quite easy to add this i hope.
Default should be `false`

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/efs.py`](https://github.com/['ansible-collections/amazon.aws', 'ansible-collections/community.aws', 'ansible-collections/community.vmware']/blob/main/plugins/modules/efs.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @akazakov @jillr @markuman @ryansydnor @s-hertel @tremble
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

akamanzi:
@markuman Can I take on this issue?

markuman:
> @markuman Can I take on this issue?

Yes, of course!

akamanzi:
@markuman @nick-zh ,

Just to confirm that I am doing addressing this issue the right way, the idea is to add a Backup parameter (which is a boolean) ```create_file_system``` function (L416) and in the ```main``` function

https://github.com/ansible-collections/community.aws/blob/cb9716e14d44357aaadd2be733bbaa0dd8a522bc/plugins/modules/efs.py#L416

markuman:
@akamanzi yes, exactly. basically you must complete four things.

1. `create_file_system`. easy
2. `update_file_system`
  * it must be possible to enable/disable the backup feature on existing EFS
  * but take care about backwards compatibility on existing playbooks. therefore, the `backup` parameter must not have a default value.
3. expand the integration test
4. add a changelog fragment to your PR

akamanzi:
@markuman , Thank you for the guidance. 
[...]

PR #1502, opened 2022-09-19
EFS: Added backup option

akamanzi:
##### SUMMARY

Added backup option to the `efs` module which enables automatic backup of the file system.
Addresses issue #1266 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Impacted component: `efs` module
##### ADDITIONAL INFORMATION

To enable automatic backup, you can add the following to the playbook
```
- name: Efs enable automatic backup 
      efs:
        state: present
        name: 'efs-name'
        tags:
            Name: 'tag-name'
        targets:
            - subnet_id: subnet-id
            - subnet_id: subnet-id
        backup: 'enabled'
```
Disabling automatic backup:

```
- name: Efs enable automatic backup 
      efs:
        state: present
        name: 'efs-name'
        tags:
            Name: 'tag-name'
        targets:
            - subnet_id: 'subnet-id'
            - subnet_id: 'subnet-id'
        backup: 'disabled'
```

```paste below

```

### DIFF
--- a/plugins/modules/efs.py
+++ b/plugins/modules/efs.py
@@ -79,6 +79,12 @@ options:
         description:
             - If the throughput_mode is provisioned, select the amount of throughput to provisioned in Mibps.
         type: float
+    backup:
+        description:
+            - Changes the file system backup policy. Use this option to start or stop automatic backups of the file system.
+            - By default, automatic backup of the file system is disabled.
+        choices: ['enabled', 'disabled']
+        type: str
     wait:
         description:
             - "In case of 'present' state should wait for EFS 'available' life cycle state (of course, if current state not 'deleting' or 'deleted')
@@ -274,6 +280,7 @@ class EFSConnection(object):
     STATE_AVAILABLE = 'available'
     STATE_DELETING = 'deleting'
     STATE_DELETED = 'deleted'
+    BACKUP_DISABLED = 'DISABLED'
 
     def __init__(self, module):
         self.connection = module.client('efs')
@@ -413,6 +420,24 @@ class EFSConnection(object):
         ))
         return info.get('ProvisionedThroughputInMibps', None)
 
+    def get_backup_policy(self, fs_id):
+        """
+         Returns the status of Filesystem backup policy
+        """
+        try:
+            info = self.connection.describe_backup_policy(FileSystemId=fs_id)
+            return info['BackupPolicy']['Status']
+        except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
+            # PolicyNotFound error returned whenever default backup policy is in effect
+            # In this case, we return disabled status
+            # By default, the automatic backup is disabled
+            if e.response['Error']['Code'] == 'PolicyNotFound':
+                return self.BACKUP_DISABLED
+            else:
+                self.module.fail_json_aws(e, msg="Unable to get file system backup policy.")
+
+        return self.BACKUP_DISABLED
+
     def create_file_system(self, name, performance_mode, encrypt, kms_key_id, throughput_mode, provisioned_throughput_in_mibps):
         """
          Creates new filesystem with selected name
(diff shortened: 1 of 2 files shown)


akamanzi:
@markuman, this feature requires `elasticfilesystem:PutBackupPolicy` and `elasticfilesystem:DescribeBackupPolicy` privileges on AWS

markuman:
@akamanzi cool to see your PR.
Permissions are handled in a different project for community.aws 

https://github.com/mattclay/aws-terminator/blob/master/aws/policy/storage-services.yaml

That means you must made a 2nd pull request there and reference to your PR here.

Unfortunately I'm out of office until end of month, maybe @alinabuzachis can assist you there.

akamanzi:
@markuman, thanks for getting back.

@alinabuzachis, I made the second PR to the `aws-terminator` (https://github.com/mattclay/aws-terminator/pull/228) adding the required policies.

plugins/modules/efs.py
```diff
@@ -79,6 +79,12 @@
         description:
             - If the throughput_mode is provisioned, select the amount of throughput to provisioned in Mibps.
         type: float
+    backup:
+        description:
+            - Changes the file system backup policy. Use this option to start or stop automatic backups of the file system.
+            - By default, automatic backup of the file system is disabled.
+        choices: ['enabled', 'disabled']
```
goneri, line 86: There is more choices available: 'ENABLED'|'ENABLING'|'DISABLED'|'DISABLING'

See: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/efs.html#EFS.Client.put_backup_policy

plugins/modules/efs.py
```diff
@@ -79,6 +79,12 @@
         description:
             - If the throughput_mode is provisioned, select the amount of throughput to provisioned in Mibps.
         type: float
+    backup:
```
goneri, line 82: Can you call it `backup_policy` to be consistent with AWS's naming.

plugins/modules/efs.py
```diff
@@ -274,6 +280,7 @@ class EFSConnection(object):
     STATE_AVAILABLE = 'available'
     STATE_DELETING = 'deleting'
     STATE_DELETED = 'deleted'
+    BACKUP_DISABLED = 'DISABLED'
```
goneri, line 283: I'm not sure the constant increase the readability.

plugins/modules/efs.py
```diff
@@ -502,6 +527,30 @@ def update_lifecycle_policy(self, name, transition_to_ia):
                 changed = True
         return changed
 
+    def change_backup_policy(self, name, status):
+        """
+        Enable/disable automatic backup policy of the FileSystem.
+        """
+        changed = False
+        state = self.get_file_system_state(name)
+        if state in [self.STATE_AVAILABLE, self.STATE_CREATING]:
```
goneri, line 536: If `state` doesn't match the condition, we silently ignore it and return "unchanged" even if the value is not the one expected.

plugins/modules/efs.py
```diff
@@ -502,6 +527,30 @@ def update_lifecycle_policy(self, name, transition_to_ia):
...
+            wait_for(
+                lambda: self.get_file_system_state(name),
+                self.STATE_AVAILABLE,
+                self.wait_timeout
+            )
+            fs_id = self.get_file_system_id(name)
+            current_backup_policy = self.get_backup_policy(fs_id)
+            params = dict()
```
goneri, line 544: The code is valide, but `params` in general is already used for the module parameters. Can you use a different name for your variable?

plugins/modules/efs.py
```diff
@@ -767,6 +818,9 @@ def main():
                                                   throughput_mode=throughput_mode, provisioned_throughput_in_mibps=provisioned_throughput_in_mibps) or changed
         if transition_to_ia:
             changed |= connection.update_lifecycle_policy(name, transition_to_ia)
+        if backup:
+            backup = str(backup).upper()
```
goneri, line 822: `backup` is already a string, you don't need the `str()` cast.

plugins/modules/efs.py
```diff
@@ -767,6 +818,9 @@ def main():
...
+            changed = connection.change_backup_policy(name, backup)
```
goneri, line 823: It should `changed |=`, not just `=`, this way we preserve the value if it's set `True` above.

e.g:
```python
>>> changed = False
>>> changed |= False
>>> changed |= True
>>> changed |= False
>>> changed
True
```

goneri (changes requested):
Thank you @akamanzi for the PR.

plugins/modules/efs.py
akamanzi: @goneri , thank you for the detailed review. 
For this part, I intentionally left out the other two choices (ENABLING & DISABLING) since when any of the two is passed as a parameter, the error `BadRequest` is returned. The message of the error states that only `ENABLED` or `DISABLED` should be specified. The error message below:

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: botocore.errorfactory.BadRequest: An error occurred (BadRequest) when calling the PutBackupPolicy operation: PutBackupPolicy must specify either 'ENABLED' or 'DISABLED'

``` 

So it seemed to me to include them as choices would only confuse users since their use would result in an error. 

Any thoughts on this situation?

plugins/modules/efs.py
akamanzi: This makes sense, I will change this

plugins/modules/efs.py
akamanzi: would `BACKUP_POLICY_DISABLED` or `BACKUP_POLICY_STATUS_DISABLED`  improve on readability?

plugins/modules/efs.py
akamanzi: Sure

plugins/modules/efs.py
akamanzi: Thanks for this catch

plugins/modules/efs.py
akamanzi: Not sure I understand what you mean by this, isn't the "unchanged" when the `changed= False`?

PR #1557, opened 2022-10-11
Add origin shield in cloudfront_distribution module

boutetnico:
##### SUMMARY
Add Origin Shield option to `cloudfront_distribution` module.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`cloudfront_distribution.py`

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/cloudfront_distribution.py
+++ b/plugins/modules/cloudfront_distribution.py
@@ -1759,12 +1759,13 @@ class CloudFrontValidationManager(object):
                 origin['custom_headers'] = ansible_list_to_cloudfront_list()
             if 'origin_shield' in origin:
                 origin_shield = origin.get('origin_shield')
-                if origin_shield.get('enabled') is not None:
-                    if origin_shield['enabled']:
-                        origin_shield_region = origin_shield.get('origin_shield_region').lower()
-                        if origin_shield_region is None:
-                            self.module.fail_json(msg="origins[].origin_shield.origin_shield_region must be specified"
+                if origin_shield.get('enabled'):
+                    origin_shield_region = origin_shield.get('origin_shield_region')
+                    if origin_shield_region is None:
+                        self.module.fail_json(msg="origins[].origin_shield.origin_shield_region must be specified"
                                                   " when origins[].origin_shield.enabled is true.")
+                    else:
+                        origin_shield_region = origin_shield_region.lower()
             if self.__s3_bucket_domain_identifier in origin.get('domain_name').lower():
                 if origin.get("s3_origin_access_identity_enabled") is not None:
                     if origin['s3_origin_access_identity_enabled']:
--- a/plugins/modules/cloudfront_distribution.py
+++ b/plugins/modules/cloudfront_distribution.py
@@ -1756,12 +1756,12 @@ class CloudFrontValidationManager(object):
             else:
                 origin['custom_headers'] = ansible_list_to_cloudfront_list()
             if 'origin_shield' in origin:
-              origin_shield = origin.get('origin_shield')
-              if origin_shield.get('enabled') is not None:
-                if origin_shield['enabled']:
-                  origin_shield_region = origin_shield.get('origin_shield_region').lower()
-                  if origin_shield_region is None:
-                      self.module.fail_json(msg="origins[].origin_shield.origin_shield_region must be specified when origins[].origin_shield.enabled is true.")
+                origin_shield = origin.get('origin_shield')
+                if origin_shield.get('enabled') is not None:
+                    if origin_shield['enabled']:
+                        origin_shield_region = origin_shield.get('origin_shield_region').lower()
+                        if origin_shield_region is None:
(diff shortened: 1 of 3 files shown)


plugins/modules/cloudfront_distribution.py
```diff
@@ -118,6 +118,16 @@
         origin_path:
           description: Tells CloudFront to request your content from a directory in your Amazon S3 bucket or your custom origin.
           type: str
+        origin_shield:
+          description: Specify origin shield options for the origin.
```
alinabuzachis, line 122: Please add `version_added: 5.1.0`

plugins/modules/cloudfront_distribution.py
```diff
@@ -1277,6 +1287,21 @@
           returned: always
           type: str
           sample: ''
+        origin_shield:
+          description: Configuration of the origin Origin Shield.
```
alinabuzachis, line 1291: Please add `version_added: 5.1.0`

alinabuzachis (commented):
@boutetnico Thank you for working on this feature. Left two minor comments. Otherwise, LGTM!

plugins/modules/cloudfront_distribution.py
```diff
@@ -1711,6 +1738,14 @@ def validate_origin(self, client, existing_config, origin, default_origin_path):
...
             else:
                 origin['custom_headers'] = ansible_list_to_cloudfront_list()
+            if 'origin_shield' in origin:
+                origin_shield = origin.get('origin_shield')
+                if origin_shield.get('enabled') is not None:
+                    if origin_shield['enabled']:
+                        origin_shield_region = origin_shield.get('origin_shield_region').lower()
+                        if origin_shield_region is None:
+                            self.module.fail_json(msg="origins[].origin_shield.origin_shield_region must be specified"
+                                                  " when origins[].origin_shield.enabled is true.")
```
markuman, line 1748:
```suggestion
            if origin.get('origin_shield', {}).get('enabled', False):
                origin_shield_region = origin.get('origin_shield').get('origin_shield_region').lower()
                if origin_shield_region is None:
                    self.module.fail_json(msg="origins[].origin_shield.origin_shield_region must be specified"
                                            " when origins[].origin_shield.enabled is true.")
```

I'm not sure if this increases the readability ....

tests/integration/targets/cloudfront_distribution/tasks/main.yml
```diff
@@ -126,6 +126,26 @@
...
+      state: present
+    register: update_origin_origin_shield
+
+  - name: ensure origin Origin Shield was enabled
+    assert:
+      that:
+        - update_origin_origin_shield.changed
+        - update_origin_origin_shield.origins.items[0].origin_shield.enabled
+        - update_origin_origin_shield.origins.items[0].origin_shield.origin_shield_region == '{{ aws_region }}'
+
```
markuman, line 148: can you also please an idempotent test and also a 3rd one that disable it again?

plugins/modules/cloudfront_distribution.py
```diff
@@ -1711,6 +1738,14 @@ def validate_origin(self, client, existing_config, origin, default_origin_path):
                 origin['custom_headers'] = ansible_list_to_cloudfront_list(origin.get('custom_headers'))
             else:
                 origin['custom_headers'] = ansible_list_to_cloudfront_list()
+            if 'origin_shield' in origin:
+                origin_shield = origin.get('origin_shield')
+                if origin_shield.get('enabled') is not None:
```
goneri, line 1743: You can replace the two first lines:
```python
if origin_shield.get('enabled') is not None:
    if origin_shield['enabled']:
```

With just:

`if origin_shield.get('enabled'):`

`.get()` will return `None` if the key is missing. Python sees both `None` and `False` as `False`.

plugins/modules/cloudfront_distribution.py
```diff
@@ -1711,6 +1738,14 @@ def validate_origin(self, client, existing_config, origin, default_origin_path):
...
+                    if origin_shield['enabled']:
+                        origin_shield_region = origin_shield.get('origin_shield_region').lower()
```
goneri, line 1745: You will get this exception if `origin_shield_region` is not set and default on the `None` type:

```
>>> None.lower()
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'lower'
```

You should do the `lower()` call later when you're sure you've got a String.

goneri (dismissed):
Looks good to me minor so tiny code adjustments.

plugins/modules/cloudfront_distribution.py
goneri: It looks better indeed. You don't need to default on False on line 1741, the default (`None`) is fine.
line 1742, you should default on a string to avoid the TypeError if `.get('origin_shield_region')` returns `None`.
line 1743, no need to compare with `None`.

markuman (approved):

alinabuzachis (approved):
@boutetnico Thank you. LGTM!

PR #1824, opened 2023-05-23
cloudfront_distribution: now honours the enabled setting

gsimon75:
##### SUMMARY

Fixes: #1823 

The `enabled: false` setting was ignored, because [here](https://github.com/ansible-collections/community.aws/blob/e80bf933412ea5c7ab2a94af945170cb2ebd900f/plugins/modules/cloudfront_distribution.py#L2221) we were falling back to the default `True` not only when the setting was `None`, but also when it was `False`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`cloudfront_distribution`

### DIFF
--- a/changelogs/fragments/1823-cloudfront_distribution_always_created_enabled.yml
+++ b/changelogs/fragments/1823-cloudfront_distribution_always_created_enabled.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - cloudfront_distribution - now honours the `enabled` setting
+  - cloudfront_distribution - now honours the ``enabled`` setting (https://github.com/ansible-collections/community.aws/issues/1823).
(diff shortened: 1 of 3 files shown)


changelogs/fragments/1823-cloudfront_distribution_always_created_enabled.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - cloudfront_distribution - now honours the `enabled` setting
```
markuman, line 2:
```suggestion
  - cloudfront_distribution - now honours the `enabled` setting (https://github.com/ansible-collections/community.aws/issues/1823).
```

markuman:
Is it possible that you add an integration test that covers this case @gsimon75?  
Otherwise looks good to me.

markuman (approved):

alinabuzachis (approved):

PR #1916, opened 2023-08-14
Add new purge_stack_instances parameter for the CloudFormation stack set module

rmahroua:
##### SUMMARY
This PR extends the `cloudformation_stack_set` module by adding a `purge_stack_instances` module parameter. Currently, the module is not idempotent because removing an account from the list of accounts doesn't instruct CloudFormation to remove account instances from the stack set. This change ensures that CloudFormation keeps stack instances only for the accounts listed in the `accounts` list.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`cloudformation_stack_set`

### DIFF
--- a/plugins/modules/cloudformation_stack_set.py
+++ b/plugins/modules/cloudformation_stack_set.py
@@ -66,14 +66,19 @@ options:
     type: str
   purge_stacks:
     description:
-    - Only applicable when I(state=absent). Sets whether, when deleting a stack set, the stack instances should also be deleted.
-    - By default, instances will be deleted. To keep stacks when stack set is deleted set I(purge_stacks=false).
+      - Only applicable when I(state=absent). Sets whether, when deleting a stack set, the stack instances should also be deleted.
+      - By default, instances will be deleted. To keep stacks when stack set is deleted set I(purge_stacks=false).
     type: bool
     default: true
+  purge_stack_instances:
+    description:
+      - When set to I(True), instructs the module to delete stack instances that are not defined via the C(accounts) parameter.
+    type: bool
+    default: false
   wait:
     description:
-    - Whether or not to wait for stack operation to complete. This includes waiting for stack instances to reach UPDATE_COMPLETE status.
-    - If you choose not to wait, this module will not notify when stack operations fail because it will not wait for them to finish.
+      - Whether or not to wait for stack operation to complete. This includes waiting for stack instances to reach UPDATE_COMPLETE status.
+      - If you choose not to wait, this module will not notify when stack operations fail because it will not wait for them to finish.
     type: bool
     default: false
   wait_timeout:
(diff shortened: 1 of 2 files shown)


plugins/modules/cloudformation_stack_set.py
```diff
@@ -83,51 +88,51 @@
...
+      - The default name for the execution role is C(AWSCloudFormationStackSetExecutionRole)
     aliases:
-    - exec_role_name
-    - exec_role
-    - execution_role
-    type: str
+      - exec_role_name
+      - exec_role
+      - execution_role
+      type: str
```
felixfontein, line 135: You also indented `type`, that turns this into invalid YAML.

Razique:
@alinabuzachis would you know how to fix the pipeline? I'm not quite sure I understand where the error is.
thank you!

plugins/modules/cloudformation_stack_set.py
```diff
@@ -66,14 +66,19 @@
...
     description:
-    - Only applicable when I(state=absent). Sets whether, when deleting a stack set, the stack instances should also be deleted.
-    - By default, instances will be deleted. To keep stacks when stack set is deleted set I(purge_stacks=false).
+      - Only applicable when I(state=absent). Sets whether, when deleting a stack set, the stack instances should also be deleted.
+      - By default, instances will be deleted. To keep stacks when stack set is deleted set I(purge_stacks=false).
     type: bool
     default: true
+  purge_stack_instances:
+    description:
+      - When set to I(True), instructs the module to delete stack instances that are not defined via the C(accounts) parameter.
```
felixfontein, line 75:
```suggestion
      - When set to C(true), instructs the module to delete stack instances that are not defined via the I(accounts) parameter.
```

plugins/modules/cloudformation_stack_set.py
```diff
@@ -83,50 +88,50 @@
...
+      - 'CAPABILITY_NAMED_IAM'
   regions:
     description:
-    - A list of AWS regions to create instances of a stack in. The I(region) parameter chooses where the Stack Set is created, and I(regions)
-      specifies the region for stack instances.
-    - At least one region must be specified to create a stack set. On updates, if fewer regions are specified only the specified regions will
+      - A list of AWS regions to create instances of a stack in. The I(region) parameter chooses where the Stack Set is created, and I(regions)
+        specifies the region for stack instances.
+      - At least one region must be specified to create a stack set. On updates, if fewer regions are specified only the specified regions will
       have their stack instances updated.
```
felixfontein, line 106: Many of the sanity test errors are because you forgot to indent this line by two spaces. This makes the `DOCUMENTATION` invalid YAML.

plugins/modules/cloudformation_stack_set.py
```diff
@@ -705,10 +730,21 @@ def main():
...
+        elif unspecified_stack_instances and module.params.get("purge_stack_instances"):
+            operation_ids.append(f"Ansible-StackInstance-RemoveInstance-{operation_uuid}")
+            changed = True
+            cfn.delete_stack_instances(
+                StackSetName=module.params["name"],
+                Accounts=list(set(acct for acct, _ in unspecified_stack_instances)),
+                Regions=list(set(region for _, region in unspecified_stack_instances)),
                 OperationPreferences=get_operation_preferences(module),
                 OperationId=operation_ids[-1],
+                RetainStacks=not(module.params.get("purge_stacks"))
```
felixfontein, line 747: The missing space after `not` is another sanity test failure. The brackets are not necessary at all, and it's a good idea to add a trailing comma here as the previous last line had.
```suggestion
                RetainStacks=not module.params.get("purge_stacks"),
```

plugins/modules/cloudformation_stack_set.py
```diff
@@ -705,10 +730,21 @@ def main():
             changed = True
             cfn.create_stack_instances(
                 StackSetName=module.params["name"],
-                Accounts=list(set(acct for acct, region in new_stack_instances)),
-                Regions=list(set(region for acct, region in new_stack_instances)),
+                Accounts=list(set(acct for acct, _ in new_stack_instances)),
```
felixfontein, line 733: Using `_` for a dummy variable is not allowed by the sanity tests. Use `dummy` instead.
```suggestion
                Accounts=list(set(acct for acct, dummy in new_stack_instances)),
```

plugins/modules/cloudformation_stack_set.py
Razique: @felixfontein I think this module predates the YAML linting check. I have fixed the YAML indentation for the whole documentation block

felixfontein: The module had valid YAML indentation before, and passed linting, before you started adjusting indentation.

rmahroua: All nested list items were at the same level as their parent, although the tests were passing, the indentation was not correct.

felixfontein: That indentation is correct as well. It's valid YAML. That's all that counts.

Just because the default config of yamllint doesn't like it doesn't mean it's not correct.

rmahroua: Fair, I should have said "inconsistent" instead of "not correct". Inconsistent because all other other modules use indentation for their documentation. The official YAML spec. also uses two spaces to indent child items (regardless of their types).

felixfontein: Yeah, they definitely were not consistent :) About the YAML spec: do you mean the examples in it?

Razique: Yeah that's what I meant. You're correct though, spaces are defined as a convention or "presentation detail" in the specs ([6.1 Identation Spaces](https://yaml.org/spec/1.2.2/#61-indentation-spaces))

PR #2104, opened 2024-06-14
fix error in module iam_server_certificate for _compare_cert

andrewjroth:
##### SUMMARY
While using the module `iam_server_certificate`, I discovered that the module does not properly compare an existing certificate with the passed value.  This happens because the `str.replace` method does not replace the variable in-place, but returns the string after replacement.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
iam_server_certificate

##### ADDITIONAL INFORMATION
To reproduce the error, run the module or include it in a playbook.  Assuming the module's arguments remain the same, the first run of the playbook will succeed.  The second run will fail with the message "Modifying the certificate body is not supported by AWS".  The expected result should be a success, not changed.

### DIFF
--- a/plugins/modules/iam_server_certificate.py
+++ b/plugins/modules/iam_server_certificate.py
@@ -137,12 +137,8 @@ def _compare_cert(cert_a, cert_b):
     # Trim out the whitespace before comparing the certs.  While this could mean
     # an invalid cert 'matches' a valid cert, that's better than some stray
     # whitespace breaking things
-    cert_a.replace("\r", "")
-    cert_a.replace("\n", "")
-    cert_a.replace(" ", "")
-    cert_b.replace("\r", "")
-    cert_b.replace("\n", "")
-    cert_b.replace(" ", "")
+    cert_a = cert_a.replace("\r", "").replace("\n", "").replace(" ", "")
+    cert_b = cert_b.replace("\r", "").replace("\n", "").replace(" ", "")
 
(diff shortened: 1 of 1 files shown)


tremble (changes requested):
Hi @andrewjroth,

Thanks for taking the time to submit this PR.  Two things,

1. To help avoid regressions, please could you add an integration test to either this PR or #2105   (tests/integration/targets/iam_server_certificate)
2. please add a changelog entry https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs-how-to

andrewjroth:
@tremble Looking at the existing integration tests, I feel that they might already cover this.

When attempting to run the integration tasks locally, I get an error...

I run:
```
$ ansible-test integration --docker
```
... and get ...
```
ERROR! couldn't resolve module/action 'community.crypto.openssl_privatekey'. This often indicates a misspelling, missing collection, or incorrect module path.

The error appears to be in '/root/ansible_collections/community/aws/tests/output/.tmp/integration/acm_certificate-tfx8pg2l-ÅÑŚÌβŁÈ/tests/integration/targets/acm_certificate/tasks/main.yml': line 51, column 5, but may
be elsewhere in the file depending on the exact syntax problem.
```

This is using commit `9c66d1e4` -- Bump release for main branch to 9.0.0-dev0 (#2091)

PR #2105, opened 2024-06-15
Add Support for Updating in iam_server_certificate by Deleting and Creating

andrewjroth:
##### SUMMARY
_This PR relies on #2104 fixing the certificate comparison. Please merge #2104 first._

Previously, changes to the certificate would fail because the module does not support modifications.  This change allows for updates by deleting the existing certificate before creating it again.  While AWS does not support updating the certificate in-place, it can be supported by delete and replace.

Because this addresses a previously failed scenario, I don't believe this will break any existing usage.  The newly created certificate will have the same name and ARN as the previous certificate, meaning it can continue to be referenced the same.  The AWS-assigned random ID number will change, but this value is returned in the result from the module so that any downstream usage of the ID will necessarily have a different value and will update accordingly.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
iam_server_certificate

### DIFF
--- a/plugins/modules/iam_server_certificate.py
+++ b/plugins/modules/iam_server_certificate.py
@@ -11,6 +11,9 @@ version_added: 1.0.0
 short_description: Manage IAM server certificates for use on ELBs and CloudFront
 description:
   - Allows for the management of IAM server certificates.
+  - If a certificate already exists matching the name, but the certificate or chain is different,
+    the certificate will be deleted and recreated.  This will result in the same
+    I(ServerCertificateName) and I(Arn), but the I(ServerCertificateId) may be different.
 options:
   name:
     description:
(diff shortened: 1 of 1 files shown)


tremble (changes requested):
Hi @andrewjroth,

1. Please could you add an integration test (tests/integration/targets/iam_server_certificate)
2. please add a changelog entry https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs-how-to
3. This feels like the type of change that could cause pain for anyone relying on "fail" behaviour, especially since the cert ID changes.  I think it might be better to add a parameter to "force" the delete/create behaviour, we can default this to false, for now with a deprecation warning that it'll switch to true once the deprecation period has passed (2 years, but as long as the deprecation is there our automation should flag it rather than expecting you to remember to come back.)

andrewjroth:
@tremble I commented on my other PR (#2104) about the integration test... I'll also add one here once I get past the errors.

Do you have an example of how to add a parameter with a deprecation?

PR #2264, opened 2025-03-07
Refactor `aws_ssm` display and verbosity-related methods

beeankha:
##### SUMMARY
Resolves [ACA-2239](https://issues.redhat.com/browse/ACA-2239)

Methods like `self._vvvv`, `self._vvv`, etc. have been refactored into a single method, including code from the `_display()` method.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`connection/aws_ssm`

### DIFF
--- a/plugins/connection/aws_ssm.py
+++ b/plugins/connection/aws_ssm.py
@@ -582,14 +582,14 @@ class Connection(ConnectionBase):
         else:
             host_args = {}
 
-        if level == 1:
-            display.v(to_text(message), **host_args)
-        elif level == 2:
-            display.vv(to_text(message), **host_args)
-        elif level == 3:
-            display.vvv(to_text(message), **host_args)
-        elif level == 4:
-            display.vvvv(to_text(message), **host_args)
+        verbosity_level = {
+            1: display.v,
+            2: display.vv,
+            3: display.vvv,
+            4: display.vvvv
+        }
+
+        verbosity_level[level](to_text(message), **host_args)
 
     def _get_bucket_endpoint(self):
         """
--- a/plugins/connection/aws_ssm.py
+++ b/plugins/connection/aws_ssm.py
@@ -553,7 +553,7 @@ class Connection(ConnectionBase):
         Displays the given message depending on the verbosity level.
 
         :param message: The message to display.
-        :param display_level: The verbosity level (1-4, default value of 1).
+        :param display_level: The verbosity level (1-4).
 
         :return: None
         """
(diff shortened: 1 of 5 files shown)


tests/unit/plugins/connection/aws_ssm/test_aws_ssm.py
```diff
@@ -307,6 +307,55 @@ def test_generate_commands(self, is_windows):
...
+        conn.host = conn.instance_id
+
+        with patch("time.time", return_value=100), patch.object(conn, "verbosity_display") as mock_display:
+            poll_gen = conn.poll("TEST", "test command")
+            # Advance generator twice to trigger the verbosity message
+            next(poll_gen)
+            next(poll_gen)
+
+            # Verify verbosity message contains remaining time
+            mock_display.assert_called_with(4, "TEST remaining: 10 second(s)")
```
beeankha, line 357: I figured it might be a good way to test `verbosity_display()` by utilizing a method that calls `verbosity_display()`, but maybe this is gonna cause more trouble than it's worth since the `poll()` method might move in the future.

beeankha:
recheck

plugins/connection/aws_ssm.py
```diff
@@ -540,35 +541,34 @@ def _initialize_ssm_client(self, region_name: Optional[str], profile_name: str)
...
-    def _vv(self, message: str) -> None:
-        self._display(display.vv, message)
 
-    def _vvv(self, message: str) -> None:
-        self._display(display.vvv, message)
+        verbosity_level = {1: display.v, 2: display.vv, 3: display.vvv, 4: display.vvvv}
 
-    def _vvvv(self, message: str) -> None:
-        self._display(display.vvvv, message)
+        verbosity_level[level](to_text(message), **host_args)
```
abikouo, line 567:
```suggestion
        if level not in verbosity_level.keys():
           level = 4
        verbosity_level[level](to_text(message), **host_args)
```

Either default to the highest level or raise an exception if the level is not part of the expected list

beeankha: Good call!

plugins/connection/aws_ssm.py
```diff
@@ -540,35 +541,34 @@ def _initialize_ssm_client(self, region_name: Optional[str], profile_name: str)
...
-        self._vvvv("SETUP BOTO3 CLIENTS: SSM")
+        self.verbosity_display(4, "SETUP BOTO3 CLIENTS: SSM")
         self._client = self._get_boto_client(
             "ssm",
             region_name=region_name,
             profile_name=profile_name,
         )
 
-    def _display(self, f: Any, message: str) -> None:
+    def verbosity_display(self, level: int, message: str) -> None:
```
gravesm, line 551: The function docs say there's a default value of 1, but that's not represented in the function signature.

beeankha: Good catch! I meant to remove that default value from the docstring.

abikouo (approved):

brahmanim (approved):

gravesm (approved):

PR #2270, opened 2025-03-25
connection/aws_ssm - create TerminalManager class and move related methods

mandar242:
##### SUMMARY

 create TerminalManager class and move related methods
[ACA-2102](https://issues.redhat.com/browse/ACA-2102)

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

connection/aws_ssm
##### ADDITIONAL INFORMATION

```paste below

```

PR #2272, opened 2025-03-26
Refactor connection/aws_ssm plugin

abikouo:
##### SUMMARY

Add new classes to manage AWS ssm session and subprocess session
Refer to: https://issues.redhat.com/browse/ACA-2100

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`connection/aws_ssm`

### DIFF
--- a/plugins/connection/aws_ssm.py
+++ b/plugins/connection/aws_ssm.py
@@ -323,20 +323,15 @@ EXAMPLES = r"""
         cmd: '/tmp/date.sh'
 """
 import getpass
-import json
 import os
-import pty
 import random
 import re
-import select
 import string
-import subprocess
 import time
 from functools import wraps
 from typing import Any
 from typing import Iterator
 from typing import List
-from typing import NoReturn
 from typing import Tuple
 from typing import TypedDict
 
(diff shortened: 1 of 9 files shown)


plugins/connection/aws_ssm.py
gravesm, line 1: Can you go back through and ensure that you aren't breaking the changes made in #2270? A few things I've seen so far:

You removed the `_session` property that's used here: https://github.com/ansible-collections/community.aws/blob/2c1e8c492b73ad34afa206f2faefff76f54190be/plugins/plugin_utils/terminalmanager.py#L81

You removed the `poll()` method used here: https://github.com/ansible-collections/community.aws/blob/2c1e8c492b73ad34afa206f2faefff76f54190be/plugins/plugin_utils/terminalmanager.py#L62

There may be other similar issues. These are pretty big bugs and it's a little concerning that none of our testing has caught these.

abikouo: @gravesm nothing broken here.
The `SessionManager` class has a `ProcessManager` instance attribute and implements a `__getattr__` method.

```python
def __getattr__(self, attr: str) -> Callable:
        if self._process_mgr and hasattr(self._process_mgr, attr):
            return getattr(self._process_mgr, attr)
        raise AttributeError(f"class SSMSessionManager has no attribute '{attr}'")
```

basically doing `self.connection.session_manager.stdin_write` is the same as `self.connection.session_manager.process_manager.stdin_write` (which is literally `_session.std.write`).
This is also valid for the `poll` method. It helps reduce code redundancy.

gravesm: Yeah, but it's not doing `self.connection.session_manager` it's doing `self.connection._session`.

abikouo: The `TerminalManager` is now doing `self.connection.session_manager`, it has been updated

gravesm: Ah, sorry, I missed that.

beeankha (approved):

gravesm (approved):

PR #2275, opened 2025-04-01
refactor connection/aws_ssm plugin

abikouo:
##### SUMMARY
Refactor S3 operation function from `connection/aws_ssm` plugin
Refer to: https://issues.redhat.com/browse/ACA-2100

##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/plugins/plugin_utils/base.py
+++ b/plugins/plugin_utils/base.py
@@ -8,7 +8,7 @@ from typing import Dict
 from typing import Optional
 
 try:
-    import boto3
+    from boto3.session import Session
     from botocore.client import Config
 except ImportError:
     pass
@@ -24,7 +24,7 @@ class AwsConnectionPluginBase:
         region_name: Optional[str] = None,
         endpoint_url: Optional[str] = None,
         config: Optional[Dict[str, Any]] = None,
-    ):
+    ) -> Any:
         """Gets a boto3 client based on the STS token"""
 
         aws_access_key_id = self.get_option("access_key_id")
@@ -40,7 +40,7 @@ class AwsConnectionPluginBase:
         profile_name = self.get_option("profile")
         if profile_name:
             session_args["profile_name"] = profile_name
-        session = boto3.session.Session(**session_args)
+        session = Session(**session_args)
         params = {}
         if endpoint_url:
             params["endpoint_url"] = endpoint_url
--- a/plugins/plugin_utils/base.py
+++ b/plugins/plugin_utils/base.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: Contributors to the Ansible project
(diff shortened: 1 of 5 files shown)


plugins/plugin_utils/base.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self) -> None:
+        pass
+
+    def _get_boto_client(
+        self,
+        service: str,
+        region_name: Optional[str] = None,
+        endpoint_url: Optional[str] = None,
+        config: Optional[Dict[str, Any]] = None,
+    ):
```
GomathiselviS, line 27: Should the return type be `BaseClient` ?

plugins/plugin_utils/s3clientmanager.py
```diff
@@ -110,20 +116,74 @@ def get_url(
...
+        :param out_path: Output path
+        :param method: The request method to use for the command (can be "get" or "put").
+
+        :returns: A tuple containing a shell command string and optional extra arguments for the PUT request.
+        """
+
+        command = None
+        url = None
+        put_args = None
+        if method == "get":
```
GomathiselviS, line 145: Is this right? I am little confused as you are using put_args and "PUT" method under this. Would you mind using the correct variable names and method values to avoid the confusion?

beeankha: Yes, this is confusing at first glance but it's correct (I was also puzzled by this when working on #2248 ). From my understanding, the reason for this apparent contradiction is that the code is handling file transfers between a local machine and a remote AWS instance via S3 as an intermediary. The [`ssm_action`](https://github.com/ansible-collections/community.aws/pull/2275/files#diff-8000d7fb9262e11565b61882567d729fe0958cd9b1a0105683d5bbe0e5b4e585L1068-R940) refers to what the user/caller wants to do (`get` a file from the remote instance or `put` a file to the remote instance), while the `method` in the commands refers to what the remote instance needs to do. 

Should we maybe add some comments to clarify why the specified methods seem to be the opposite of what they should be?

GomathiselviS: Thanks for the clarification, @beeankha. Yes, comments would be helpful for posterity.

abikouo: @beeankha, you are right!! The file transfer generates 2 commands, one for the controller node and another for the managed node. The controller node uses the `AWS S3` API to download/put file to S3 bucket while the managed node uses the `curl` command to download/upload file to bucket.

plugins/plugin_utils/base.py
abikouo: The botocore does use type hint, it is hard to deduced so putting `Any` as returned value is enough

plugins/plugin_utils/s3clientmanager.py
abikouo: The function description has been updated

beeankha (approved):

PR #2278, opened 2025-04-08
connection/aws_ssm - add caching and caching_ttl options to improve velocity

abikouo:
##### SUMMARY

The new options `ansible_aws_ssm_caching` and `ansible_aws_ssm_caching_ttl` will help improve the velocity of the `connection/aws_ssm` plugin.
When `ansible_aws_ssm_caching=True`, the connection plugin will start a daemon with an SSM session that will be used for the allocated `ansible_aws_ssm_caching_ttl` time to handle all commands to the EC2 managed node.
This avoids setting a new session (along with the prepare_terminal steps) for each new command.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`connection/aws_ssm`

### DIFF
--- a/plugins/connection/aws_ssm.py
+++ b/plugins/connection/aws_ssm.py
@@ -148,6 +148,23 @@ options:
     version_added: 5.2.0
     vars:
     - name: ansible_aws_ssm_s3_addressing_style
+  caching:
+    description:
+    - The plugin will create a daemon starting a shell session to handle all command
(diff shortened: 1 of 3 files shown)


acormier-maia:
How is cache_server.py used here?  Any thoughts on what we can do now that aws_ssm is moved to amazon.aws namespace?

PR #2325, opened 2025-07-09
[WIP] Add the identitystore_user module

vonschultz:
Add the community.aws.identitystore_user module, for managing users in an identity store (used by the IAM Identity Center).

##### SUMMARY

Ansible is currently lacking a way to interact with the AWS IdentityStore API, https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/identitystore.html. This change adds support for adding and removing users, and setting the most important attributes on those users. This could potentially be extended in the future to include things like group creation, group membership and more user attributes.

There is currently no way to create an identity store in Ansible, so I made the integration tests "disabled", since they rely on that external resource. You can create it in the IAM Identity Center, and maybe the `AWS::SSO::Instance` CloudFormation resource also produces identity stores that could be used for this (at most one per account, according to the docs).

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

identitystore_user

### DIFF
--- a/plugins/modules/identitystore_user.py
+++ b/plugins/modules/identitystore_user.py
@@ -1,0 +1,314 @@
+#!/usr/bin/python
+
+# Copyright (c) 2017 Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r"""
+---
+module: identitystore_user
+version_added: 0.0.0
+short_description: Manage users in Identity Store (IAM Identity Center)
+description:
+  - Manage users in the Identity Store service.
+  - The Identity Store service is used by IAM Identity Center.
+author:
+  - Christian von Schultz (@vonschultz)
+options:
+  identity_store_id:
+    description:
+      - The ID of the identity store.
+      - Found on the Settings page in IAM Identity Center.
+    type: str
+    required: true
+  display_name:
+    description:
+      - A string containing the name of the user.
+      - https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_CreateUser.html
+    type: str
+  user_name:
+    description:
+      - A string containing the user name of the user.
+    required: true
+    type: str
+  name:
+    description:
+      - A dictionary containing the full name of the user.
+      - https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_Name.html
+    type: dict
+    suboptions:
+      family_name:
+        description:
+          - The family name of the user
+        type: str
+      given_name:
+        description:
+          - The given name of the user
+        type: str
+  emails:
+    description:
+      - A list of email dictionaries containing email addresses for the user.
+      - https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_Email.html
+    type: list
+    elements: dict
+    suboptions:
+      value:
+        description:
+          - The e-mail address
+        required: true
+        type: str
+      type:
+        description:
+          - The type of e-mail address (e.g., "work")
+        required: true
+        type: str
+      primary:
+        description:
+          - Whether this e-mail address is the user's primary address
+        required: true
+        type: bool
+  state:
+    description:
+      - Create/Update or Delete a user
+    choices: [ 'present', 'absent' ]
+    required: True
+    type: str
+extends_documentation_fragment:
+  - amazon.aws.boto3
+  - amazon.aws.common.modules
+  - amazon.aws.region.modules
+"""
+
+EXAMPLES = r"""
+- name: Create/update user
+  community.aws.identitystore_user:
+    identity_store_id: d-cba0987654
+    state: present
+    user_name: john.doe@example.com
+    name:
+      family_name: Doe
+      given_name: John
+    display_name: John Doe
+    emails:
+      - value: john.doe@example.com
+        type: work
+        primary: true
+
+- name: Remove user
+  community.aws.identitystore_user:
+    identity_store_id: d-c36714741a
+    state: absent
+    user_name: john.doe@example.com
+"""
+
+RETURN = r"""
+identity_store_id:
+  description:
+    - just echos the identity_store_id
+  type: str
+  returned: success
+user_id:
+  description:
+    - the identifier for a user in the identity store
+    - only present if the user exists when we return
+  type: str
+  returned: success
+"""
+
+from ansible.module_utils.common.dict_transformations import (
+    snake_dict_to_camel_dict,
+)
+from ansible_collections.amazon.aws.plugins.module_utils.botocore import (
+    is_boto3_error_code,
+)
+from ansible_collections.amazon.aws.plugins.module_utils.exceptions import (
+    AnsibleAWSError,
+)
+from ansible_collections.amazon.aws.plugins.module_utils.modules import (
+    AnsibleAWSModule,
+)
+
+
+def flatten_dict(nested_dict):
+    flat = {}
+    for key, value in nested_dict.items():
+        if isinstance(value, dict):
+            for inner_key, inner_value in flatten_dict(value).items():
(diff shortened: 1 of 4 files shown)


plugins/modules/identitystore_user.py
```diff
@@ -0,0 +1,290 @@
+#!/usr/bin/python
+
+# Copyright (c) 2017 Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r"""
+---
+module: identitystore_user
+version_added: 0.0.0
```
alinabuzachis:
```suggestion
version_added: 10.1.0
```

plugins/modules/identitystore_user.py
```diff
@@ -0,0 +1,290 @@
...
+          - The family name of the user
+        type: str
+      given_name:
+        description:
+          - The given name of the user
+        type: str
+  emails:
+    description:
+      - A list of email dictionaries containing email addresses for the user.
+      - https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_Email.html
```
alinabuzachis:
```suggestion
      - U(https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_Email.html).
```

plugins/modules/identitystore_user.py
```diff
@@ -0,0 +1,290 @@
...
+      - https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_Name.html
+    type: dict
+    suboptions:
+      family_name:
+        description:
+          - The family name of the user
+        type: str
+      given_name:
+        description:
+          - The given name of the user
```
alinabuzachis:
```suggestion
          - The given name of the user.
```

plugins/modules/identitystore_user.py
```diff
@@ -0,0 +1,290 @@
...
+    type: str
+  name:
+    description:
+      - A dictionary containing the full name of the user.
+      - https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_Name.html
+    type: dict
+    suboptions:
+      family_name:
+        description:
+          - The family name of the user
```
alinabuzachis:
```suggestion
          - The family name of the user.
```

plugins/modules/identitystore_user.py
```diff
@@ -0,0 +1,290 @@
...
+    type: str
+  user_name:
+    description:
+      - A string containing the user name of the user.
+    required: true
+    type: str
+  name:
+    description:
+      - A dictionary containing the full name of the user.
+      - https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_Name.html
```
alinabuzachis:
```suggestion
      - U(https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_Name.html).
```

plugins/modules/identitystore_user.py
```diff
@@ -0,0 +1,290 @@
...
+  identity_store_id:
+    description:
+      - The ID of the identity store.
+      - Found on the Settings page in IAM Identity Center.
+    type: str
+    required: true
+  display_name:
+    description:
+      - A string containing the name of the user.
+      - https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_CreateUser.html
```
alinabuzachis:
```suggestion
      - U(https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_CreateUser.html).
```

plugins/modules/identitystore_user.py
```diff
@@ -0,0 +1,290 @@
...
+  emails:
+    description:
+      - A list of email dictionaries containing email addresses for the user.
+      - https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_Email.html
+    type: list
+    elements: dict
+    suboptions:
+      value:
+        description:
+          - The e-mail address
```
alinabuzachis:
```suggestion
          - The e-mail address.
```

plugins/modules/identitystore_user.py
```diff
@@ -0,0 +1,290 @@
...
+        required: true
+        type: str
+      type:
+        description:
+          - The type of e-mail address (e.g., "work")
```
alinabuzachis:
```suggestion
          - The type of e-mail address (e.g., "work").
```

plugins/modules/identitystore_user.py
```diff
@@ -0,0 +1,290 @@
...
+        required: true
+        type: str
+      primary:
+        description:
+          - Whether this e-mail address is the user's primary address
```
alinabuzachis:
```suggestion
          - Whether this e-mail address is the user's primary address.
```

plugins/modules/identitystore_user.py
```diff
@@ -0,0 +1,290 @@
...
+        required: true
+        type: bool
+  state:
+    description:
+      - Create/Update or Delete a user
```
alinabuzachis:
```suggestion
      - Create/Update or Delete a user.
```

plugins/modules/identitystore_user.py
```diff
@@ -0,0 +1,290 @@
...
+  community.aws.identitystore_user:
+    identity_store_id: d-c36714741a
+    state: absent
+    user_name: john.doe@example.com
+"""
+
+RETURN = r"""
+identity_store_id:
+  description:
+    - just echos the identity_store_id
```
alinabuzachis:
```suggestion
    - Just echos the identity_store_id.
```

plugins/modules/identitystore_user.py
```diff
@@ -0,0 +1,290 @@
...
+  type: str
+  returned: success
+user_id:
+  description:
+    - the identifier for a user in the identity store
+    - only present if the user exists when we return
```
alinabuzachis:
```suggestion
    - The identifier for a user in the identity store.
    - Only present if the user exists when we return.
```

tests/integration/targets/identitystore/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
+- name: Run identitystore tests
+  module_defaults:
+    community.aws.identitystore_user:
+      access_key: "{{ aws_access_key }}"
+      secret_key: "{{ aws_secret_key }}"
+      session_token: "{{ security_token | default(omit) }}"
+      region: "{{ aws_region }}"
```
alinabuzachis, line 7: @vonschultz Thank you for this contribution. Since there isn’t currently a module available to retrieve the Identity Store ID, I believe it would be better to add a task that uses the AWS CLI to list all instances and then extract the identity_store_id from the output of that task. I'm not familiar with this service, but, what do you think about this suggestion?

vonschultz: If an identity store was created specifically for the test, certainly, then we should try to find its ID automatically. Currently, there's no way to create an identity store in Ansible (except maybe through CloudFormation?), and furthermore, I believe it's a global resource where you can only have one for a given account. Now, here's the rub: The user running the test might very well be using the IAM Identity Center to manage logins to their AWS account, which means that we'd be creating real users in a real identity store (and if the test fails during development of some feature, potentially even leaving them in there). This might surprise the user running the test, and possibly has some real security consequences. How do we ensure that the user running the test is OK with users being created in their identity store? How do we ensure that the user knows where they should go to clean up things manually if things go awry? It seemed like the natural way to do that would be to let the user themselves specify the `identity_store_id` when running the test — that way, the user has actively consented to us doing tests with that identity store (and should things go awry, the user knows where to find the identity store and see if there's anything that shouldn't be there).

By the way, I think it would be awesome if we could run the test in the CI for community.aws, but for that to happen, someone would have to set up an identity store we can use. The tests are pretty fast, and I only disabled them because they need external resources (in the form of an identity store that can't be created from Ansible at the time of writing).

PR #2326, opened 2025-07-17
[DO NOT MERGE] Test docs build

felixfontein:
##### SUMMARY
https://github.com/ansible-community/github-docs-build/pull/102 should have fixed the docs workflows in this repo, but unfortunately restarting the workflows does not use the new code (only for the shared workflow, but not for the action from the same repo that the workflow uses :facepalm:), so I'm trying a test PR :)

##### ISSUE TYPE
- Docs Pull Request


felixfontein:
The docs build is working again - well almost, it now breaks due to some bad changelog entries. I'll create a PR to fix that.

felixfontein:
#2327 should fix this.

PR #2347, opened 2025-10-15
Add tags to ecs cluster

Shaps:
##### SUMMARY
Adds 2 new options to the `ecs_cluster` module:
- `tags` Allows adding or updating tags on an ECS Cluster. Tags are **not** replaced by default
- `purge_tags` allows re-setting the tags to the passed values. Overrides the default behaviour of not re-setting the tags.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ecs_cluster

##### ADDITIONAL INFORMATION
Not much additional info to be added, I required the tags to be added on cluster creation.

I have also added some lint fixes in the integration test playbook/tasks

### DIFF
--- a/changelogs/fragments/20251015_ecs_cluster_add_tags.yml
+++ b/changelogs/fragments/20251015_ecs_cluster_add_tags.yml
@@ -1,3 +1,3 @@
 minor_changes:
-  - ecs_cluster - Add `tags` option to add or update tags to a cluster
-  - ecs_cluster - Add `purge_tags` option to remove tags from a cluster
+  - ecs_cluster - Add ``tags`` option to add or update tags to a cluster (https://github.com/ansible-collections/community.aws/pull/2347).
+  - ecs_cluster - Add `purge_tags` option to remove tags from a cluster (https://github.com/ansible-collections/community.aws/pull/2347).


plugins/modules/ecs_cluster.py
```diff
@@ -76,9 +76,23 @@
         required: false
         type: bool
         default: true
+    purge_tags:
+        version_added: 10.1.0
```
alinabuzachis: There is no need to manually document tags and purge_tags here, as they are automatically included when the amazon.aws.tags doc fragment is imported. However, I suggest adding a note for the module to indicate when support for these parameters was introduced (likely version 11.0.0).

plugins/modules/ecs_cluster.py
```diff
@@ -161,6 +175,11 @@
     returned: always
     type: str
     sample: ACTIVE
+tags:
+    description: Cluster tags
```
alinabuzachis:
```suggestion
    description: Cluster tags.
```

changelogs/fragments/20251015_ecs_cluster_add_tags.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - ecs_cluster - Add `tags` option to add or update tags to a cluster
```
alinabuzachis:
```suggestion
  - ecs_cluster - Add ``tags`` option to add or update tags to a cluster (https://github.com/ansible-collections/community.aws/pull/2347).
```

changelogs/fragments/20251015_ecs_cluster_add_tags.yml
```diff
@@ -0,0 +1,3 @@
...
+  - ecs_cluster - Add `purge_tags` option to remove tags from a cluster
```
alinabuzachis:
```suggestion
  - ecs_cluster - Add ``purge_tags`` option to remove tags from a cluster (https://github.com/ansible-collections/community.aws/pull/2347).
```

plugins/modules/ecs_cluster.py
Shaps: thanks. I will update the version once we're ok merging this.

plugins/modules/ecs_cluster.py
```diff
@@ -249,6 +283,8 @@ def main():
                 base=dict(type="int", default=0),
             ),
         ),
+        tags=dict(type="dict", required=False, aliases=["resource_tags"]),
+        purge_tags=dict(required=False, type="bool", default=False),
```
s-hertel, line 287: This should default to True. It should only purge existing tags if tags is explicitly set to a dictionary.

plugins/modules/ecs_cluster.py
```diff
@@ -295,6 +334,23 @@ def main():
                 requested_cp = existing_cp
                 requested_cps = existing_cps
 
+            if module.params["tags"]:
```
s-hertel, line 337: If it's an empty dictionary I'd expect it to remove all tags, not skip purging.

plugins/modules/ecs_cluster.py
```diff
@@ -295,6 +334,23 @@ def main():
...
+                else:
+                    tags_to_remove = None
+
+                cluster_mgr.update_tags(
+                    existing["clusterArn"],
+                    tags_to_remove,
+                    module.params["tags"],
+                )
+
+                tags_changed = True if recursive_diff(existing_tags, module.params["tags"]) is not None else False
```
s-hertel, line 350: Why did you use recursive_diff instead the compare_aws_tags helper? There are some general guidelines here https://docs.ansible.com/projects/ansible/latest/collections/amazon/aws/docsite/dev_guidelines.html#dealing-with-tags.

plugins/modules/ecs_cluster.py
```diff
@@ -295,6 +334,23 @@ def main():
                 requested_cp = existing_cp
                 requested_cps = existing_cps
 
+            if module.params["tags"]:
+                if module.params["purge_tags"]:
+                    tags_to_remove = existing_tags
```
s-hertel, line 339: I'd expect tags_to_remove only contains existing tags that are not specified for the task.

tests/integration/targets/ecs_cluster/defaults/main.yml
```diff
@@ -12,17 +12,17 @@ ecs_task_memory: 128
 target_swap_mb: 0
 target_swappiness: 80
 ecs_task_containers:
-- name: "{{ ecs_task_name }}"
```
s-hertel, line 15: I recommend removing all the stylistic changes to existing tests (indentation, booleans, quotes), unless maintainers ask for such a change. This will make your pull request easier to review and avoids polluting the git blame with unenforced personal preferences that can easily be changed again by the next contributor.

plugins/modules/ecs_cluster.py
```diff
@@ -304,15 +360,19 @@ def main():
                         capacity_provider_strategy=requested_cps,
                     )
                 results["changed"] = True
+
             else:
                 results["cluster"] = existing
+
+            results["cluster"] = cluster_mgr.refresh_cluster_state(cluster_name=module.params["name"])
```
s-hertel, line 367: I think you could just set the cluster tags to the calculated value. Calling describe_clusters again seems unnecessary.

plugins/modules/ecs_cluster.py
```diff
@@ -295,6 +334,23 @@ def main():
...
+                    tags_to_remove = existing_tags
+                    tags_changed = True
+                else:
+                    tags_to_remove = None
+
+                cluster_mgr.update_tags(
+                    existing["clusterArn"],
+                    tags_to_remove,
+                    module.params["tags"],
+                )
```
s-hertel, line 348: One more thought - this modules supports check mode, and this breaks it. You should add a test to make sure check mode detects tag changes but doesn't actually modify tags.

PR #2408, opened 2026-02-07
ecs_service: add Service Connect support and wait-on-present

### LINKED ISSUE #1755: Add Service Connect support to ecs_service module
### Summary

Add Service Connect support to ecs_service module.

### Issue Type

Feature Idea

### Component Name

ecs_service.py

### Additional Information

    .....
   "serviceConnectConfiguration": { 
      "enabled": boolean,
      "logConfiguration": { 
         "logDriver": "string",
         "options": { 
            "string" : "string" 
         },
         "secretOptions": [ 
            { 
               "name": "string",
               "valueFrom": "string"
            }
         ]
      },
      "namespace": "string",
      .....

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/service-connect.html
https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CreateService.html
https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_UpdateService.html

### Code of Conduct

jonpspri:
## Summary

Closes #1755

- Add `service_connect_configuration` parameter to `ecs_service` module for creating and updating ECS services with AWS Service Connect
- Extend `wait` parameter to support `state=present` using the `services_stable` waiter, blocking until `runningCount` matches `desiredCount`
- Convert positional args to keyword args at `create_service` and `update_service` call sites to prevent silent breakage on parameter reordering
- Extract `is_dict_subset()` into `plugins/module_utils/dict.py` as a reusable recursive subset comparison utility with order-insensitive list matching via `list_sort_key`
- Idempotency comparison correctly reads `serviceConnectConfiguration` from the primary deployment (where AWS returns it, not the top-level Service object) and tolerates AWS-populated defaults like `discoveryName`

## Test plan

- [x] Unit tests for `is_dict_subset` (36 tests covering basic subset, nested dicts, list comparison, sort key, scalars, and real-world ECS Service Connect scenarios)
- [x] Integration tests for Service Connect: create, idempotency, update, disable, delete
- [x] Integration test teardown covers Cloud Map namespace, task definition, and service cleanup

### DIFF
--- a/tests/unit/plugins/module_utils/test_dict.py
+++ b/tests/unit/plugins/module_utils/test_dict.py
@@ -5,7 +5,6 @@ import pytest
 
 from ansible_collections.community.aws.plugins.module_utils.dict import is_dict_subset
 
-
 # ---------------------------------------------------------------------------
 # Basic dict subset checks
 # ---------------------------------------------------------------------------


tests/integration/targets/ecs_cluster/tasks/99_terminate_everything.yml
```diff
@@ -228,6 +237,14 @@
     state: absent
   ignore_errors: true
 
+- name: remove Service Connect ecs task definition
+  ecs_taskdefinition:
+    containers: "{{ ecs_service_connect_task_containers }}"
+    family: "{{ ecs_task_name }}-sc"
+    revision: "{{ ecs_sc_task_definition.taskdefinition.revision }}"
+    state: absent
+  ignore_errors: true
```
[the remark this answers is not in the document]
jonpspri: Fixed — added `when` guard for `ecs_sc_task_definition is defined`.

plugins/modules/ecs_service.py
```diff
@@ -786,6 +848,21 @@ def is_matching_service(self, expected, existing):
...
+        # not at the top-level Service.  Use the primary (first) deployment.
+        # Use a subset check because AWS may return additional default fields
+        # (e.g. discoveryName in services) that the user did not specify.
+        expected_service_connect = (
+            snake_dict_to_camel_dict(expected["service_connect_configuration"])
+            if expected.get("service_connect_configuration")
+            else {}
+        )
+        primary_deployment = (existing.get("deployments") or [{}])[0]
+        existing_service_connect = primary_deployment.get("serviceConnectConfiguration", {})
```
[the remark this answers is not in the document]
jonpspri: Fixed — when `expected.enabled == false` and existing is empty (AWS omits the key), we now treat it as `{"enabled": false}` so the subset check converges.

jonpspri:
Also note -- the change to semantics of `wait: true` for `status: present` _could_ be considered a breaking change and need some advance notice.  (The behavior is currently undefined.). Please let me know if we need to go through a process or want to take a different approach to this function.

plugins/modules/ecs_service.py
```diff
@@ -786,6 +848,26 @@ def is_matching_service(self, expected, existing):
...
+        # When Service Connect is disabled, AWS omits serviceConnectConfiguration
+        # entirely from the deployment.  Treat a missing configuration as {"enabled": False}
+        # so that an explicit "enabled: false" in expected remains idempotent.
+        if expected_service_connect.get("enabled") is False and not existing_service_connect:
+            existing_service_connect = {"enabled": False}
+        if not is_dict_subset(expected_service_connect, existing_service_connect, list_sort_key="portName"):
```
alinabuzachis, line 874: Doesn't make sense to recursive_diff(dict1, dict2): https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/common/dict_transformations.py instead so that you do not have to create the util function?

plugins/modules/ecs_service.py
```diff
@@ -268,6 +268,40 @@
...
+        description:
+          - The configuration for this service to discover and connect to services, and be discovered by, and connected from, other services within a namespace.
+          - See U(https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ServiceConnectConfiguration.html) for more details.
+          - When determining whether the service needs updating, only the fields explicitly provided are compared
+            against the existing configuration. AWS may populate additional default fields (such as C(discoveryName))
+            which are ignored for idempotency purposes. If you need to enforce a specific value for an AWS-defaulted
+            field, include it explicitly in this configuration.
+        type: dict
+        required: false
+        version_added: 7.2.0
```
alinabuzachis:
```suggestion
        version_added: 11.1.0
```

plugins/modules/ecs_service.py
```diff
@@ -383,6 +418,32 @@
...
+    service_connect_configuration:
+      enabled: true
+      namespace: my-cloudmap-namespace
+      services:
+        - port_name: http
+          client_aliases:
+            - port: 80
+
+# Create service and wait for tasks to be running
+- community.aws.ecs_service:
```
alinabuzachis:
```suggestion
 - name: Create service and wait for tasks to be running
   community.aws.ecs_service:
```

plugins/modules/ecs_service.py
```diff
@@ -383,6 +418,32 @@
       Firstname: jane
       lastName: doe
     propagate_tags: SERVICE
+
+# With Service Connect configuration
+- community.aws.ecs_service:
```
alinabuzachis:
```suggestion
- name: Create ECS Service with Service Connect configuration
  community.aws.ecs_service:
```

plugins/modules/ecs_service.py
jonpspri: Thanks for the suggestion! I looked into `recursive_diff` but it's not a fit for this use case — the core issue is **subset checking vs equality checking**.

`recursive_diff` finds differences in **both directions**. Here we need to check whether the user's expected config is a **subset** of the AWS response, ignoring extra fields AWS adds. AWS populates many default fields the user didn't specify — for example, when the user provides:

```yaml
services:
  - port_name: http
    client_aliases:
      - port: 80
```

AWS returns:

```json
{"portName": "http", "discoveryName": "http", "clientAliases": [{"port": 80, "dnsName": "http.my-ns"}]}
```

With `recursive_diff`, the extra `discoveryName` and `dnsName` would register as differences, so every run would report `changed=true` — breaking idempotency.

`is_dict_subset` also handles order-insensitive list-of-dict comparison (via `list_sort_key`), which `recursive_diff` doesn't support — it doesn't recurse into lists at all.

This same subset-matching pattern is needed elsewhere in `is_matching_service` (and potentially other modules comparing user params against AWS responses), so a small reusable utility seemed like the right approach.

plugins/module_utils/dict.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+
+def is_dict_subset(expected, existing, list_sort_key=None):
```
abikouo:
```suggestion
def is_dict_subset(expected:  Union[dict, list[dict]], existing:  Union[dict, list[dict]], list_sort_key=None) -> bool:
```
I am suggesting type formatting to better understand the scope of the function

plugins/module_utils/dict.py
```diff
@@ -0,0 +1,39 @@
...
+        if (
+            list_sort_key
+            and exp
+            and all(isinstance(i, dict) and list_sort_key in i for i in exp)
+            and all(isinstance(i, dict) and list_sort_key in i for i in ext)
+        ):
+            exp = sorted(exp, key=lambda d: d[list_sort_key])
+            ext = sorted(ext, key=lambda d: d[list_sort_key])
+        return all(is_dict_subset(e, a, list_sort_key=list_sort_key) for e, a in zip(exp, ext))
+    return expected == existing
```
abikouo, line 59:
```suggestion
    raise ValueError("...")
```
This function is supposed to compare either dict or list of dict, should then raise an exception with the type of the input var is not compliant

plugins/module_utils/dict.py
jonpspri: Done — added type annotations and also added `from __future__ import annotations` for forward-compatible syntax. Used `Union[dict, list[dict]]` for both parameters and `-> bool` return type.

plugins/module_utils/dict.py
jonpspri: Good catch — the public `is_dict_subset()` now validates that both top-level arguments are `dict` or `list` and raises `ValueError` if not. The recursive logic is extracted into a private `_is_subset()` helper so that scalar leaf comparisons (needed during dict recursion) still work correctly. Unit tests updated accordingly.

jonpspri:
Now waiting on security updates in https://github.com/mattclay/aws-terminator/pull/330 .

alinabuzachis (approved):

PR #2413, opened 2026-02-09
Handle deprecation version errors

alinabuzachis:
##### SUMMARY

Handle deprecation version errors

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


tremble:
Do you mean this sanity test failure?

```
meta/runtime.yml:0:0: The deprecation removal_version ('12.0.0') must be after the current version (SemanticVersion('12.0.0')) for dictionary value @ data['plugin_routing']['modules']['waf_condition']['deprecation']['removal_version']. Got '12.0.0'
```

Easiest option: remove the modules from main.

The reason we bump the version on main relatively early is to trigger these warnings as early as possible and encourage the breaking changes early in the release cycle.  There were a number of occasions early on where we only caught planned-but-missing breaking changes at the very end of the release cycle, which reduced the bake-in time and then caused problems with the release (where we trigger all tests)

PR #2427, opened 2026-04-12
wafv2_web_acl - Fix idempotency issue with rate_based_statement

tremble:
##### SUMMARY
Fixed idempotency issue where rules with rate_based_statement would always show as changed when evaluation_window_sec was not explicitly specified. AWS returns the default value of 300, but the module was not applying this default before comparison. Also updated documentation and test infrastructure.

##### ISSUE TYPE
Bugfix Pull Request

##### COMPONENT NAME
wafv2_web_acl

##### ADDITIONAL INFORMATION
- Added apply_rate_based_statement_defaults() function to apply AWS default values before comparison
- Added comprehensive unit tests for wafv2 module_utils functions
- Added integration test for ByteMatch and RateBased statement idempotency
- Updated RETURN documentation for read-only web ACL fields
- Updated PARAMETERS documentation to include rules.override_action which was automatically supported when the API was updated.
- Synced tox.ini and pyproject.toml configuration with amazon.aws collection

Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/plugins/modules/wafv2_web_acl.py
+++ b/plugins/modules/wafv2_web_acl.py
@@ -73,15 +73,26 @@ options:
           type: int
         action:
           description:
-            - Wether a rule is blocked, allowed or counted.
+            - Whether a rule is blocked, allowed, counted, uses CAPTCHA, or uses Challenge.
+            - Mutually exclusive with O(rules.override_action).
+            - Required for normal rules. Not used for managed rule group reference statements.
+          type: dict
+        override_action:
+          description:
+            - Override action for rule group reference statements.
+            - Can override the rule group evaluation result.
+            - Mutually exclusive with O(rules.action).
+            - Required when using managed rule group statements.
           type: dict
         visibility_config:
           description:
             - Visibility of single wafv2 rule.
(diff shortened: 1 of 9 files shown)


plugins/modules/wafv2_web_acl.py
```diff
@@ -73,15 +73,26 @@
           type: int
         action:
           description:
-            - Wether a rule is blocked, allowed or counted.
+            - Whether a rule is blocked, allowed, counted, uses CAPTCHA, or uses Challenge.
+            - Mutually exclusive with O(rules.override_action).
+            - Required for normal rules. Not used for managed rule group reference statements.
+          type: dict
+        override_action:
```
alinabuzachis, line 80: Do we need version_added for this option? Should this one be documented with minor_changes too?

alinabuzachis (approved):

plugins/modules/wafv2_web_acl.py
tremble: It's not actually new, it was just missing from the documentation

